PR #16, opened 2024-01-17
✨(api) search users by email

mjeammet:
## Purpose

We want to create the api endpoint to later allow user search by email. 

## Proposal

Creation of the api endpoint /users/ 

- [x] add `list`method to user viewset 
- [x] filters results by similarity with the query
- [x] add tests for partial, accentuated and capitalized queries
- [x] fixed tests impacted by the creation of the endpoint
- [x] decreased similarity threshold to 0.01

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -172,7 +172,9 @@ class ContactViewSet(
         return super().perform_create(serializer)
 
 
-class UserViewSet(mixins.UpdateModelMixin, viewsets.GenericViewSet):
+class UserViewSet(
+    mixins.UpdateModelMixin, viewsets.GenericViewSet, mixins.ListModelMixin
+):
     """
     User viewset for all interactions with user infos and teams.
 
@@ -186,30 +188,31 @@ class UserViewSet(mixins.UpdateModelMixin, viewsets.GenericViewSet):
     queryset = models.User.objects.all().select_related("profile_contact")
     serializer_class = serializers.UserSerializer
     throttle_classes = [BurstRateThrottle, SustainedRateThrottle]
+    pagination_class = Pagination
 
-    def list(self, request, *args, **kwargs):
-        """Limit listed users by a query with throttle protection."""
-        queryset = self.filter_queryset(self.get_queryset())
+    def get_queryset(self):
+        """Limit listed users by a query. Pagination and throttle protection apply."""
+        queryset = self.queryset
 
-        # Exclude inactive contacts
-        queryset = queryset.filter(
-            is_active=True,
-        )
-
-        # Search by case-insensitive and accent-insensitive trigram similarity
-        if query := self.request.GET.get("q", ""):
-            similarity = TrigramSimilarity(
-                Func("email", function="unaccent"),
-                Func(Value(query), function="unaccent"),
-            )
-            queryset = (
-                queryset.annotate(similarity=similarity)
-                .filter(similarity__gte=EMAIL_SIMILARITY_THRESHOLD)
-                .order_by("-similarity")
+        if self.action == "list":
+            # Exclude inactive contacts
+            queryset = queryset.filter(
+                is_active=True,
             )
 
-        serializer = self.get_serializer(queryset, many=True)
-        return response.Response(serializer.data)
+            # Search by case-insensitive and accent-insensitive trigram similarity
+            if query := self.request.GET.get("q", ""):
+                similarity = TrigramSimilarity(
+                    Func("email", function="unaccent"),
+                    Func(Value(query), function="unaccent"),
+                )
+                queryset = (
+                    queryset.annotate(similarity=similarity)
+                    .filter(similarity__gte=EMAIL_SIMILARITY_THRESHOLD)
+                    .order_by("-similarity")
+                )
+
+        return queryset
 
     @decorators.action(
         detail=False,
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -116,9 +116,6 @@ class ContactViewSet(
         user = self.request.user
         queryset = self.filter_queryset(self.get_queryset())
 
-        if not user.is_authenticated:
-            return queryset.none()
-
         # Exclude contacts that:
         queryset = queryset.filter(
             # - belong to another user (keep public and owned contacts)
@@ -240,7 +237,6 @@ class UserViewSet(mixins.UpdateModelMixin, viewsets.GenericViewSet):
         methods=["get"],
         url_name="me",
         url_path="me",
-        permission_classes=[permissions.IsAuthenticated],
     )
     def get_me(self, request):
         """
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1,6 +1,5 @@
 """API endpoints"""
 from django.contrib.postgres.search import TrigramSimilarity
-from django.core.cache import cache
 from django.db.models import Func, OuterRef, Q, Subquery, Value
 
 from rest_framework import (
@@ -11,6 +10,7 @@ from rest_framework import (
     response,
     viewsets,
 )
+from rest_framework.throttling import UserRateThrottle
 
 from core import models
 
@@ -97,6 +97,22 @@ class Pagination(pagination.PageNumberPagination):
     page_size_query_param = "page_size"
 
 
+class BurstRateThrottle(UserRateThrottle):
+    """
+    Throttle rate for minutes. See DRF section in settings for default value.
+    """
+
+    scope = "burst"
+
+
+class SustainedRateThrottle(UserRateThrottle):
+    """
+    Throttle rate for hours. See DRF section in settings for default value.
+    """
+
+    scope = "sustained"
+
+
 # pylint: disable=too-many-ancestors
 class ContactViewSet(
     mixins.CreateModelMixin,
@@ -110,6 +126,7 @@ class ContactViewSet(
     permission_classes = [permissions.IsOwnedOrPublic]
     queryset = models.Contact.objects.all()
     serializer_class = serializers.ContactSerializer
+    throttle_classes = [BurstRateThrottle, SustainedRateThrottle]
 
     def list(self, request, *args, **kwargs):
         """Limit listed users by a query with throttle protection."""
@@ -141,26 +158,6 @@ class ContactViewSet(
                 .order_by("-similarity")
             )
 
-        # Throttle protection
-        key_base = f"throttle-contact-list-{user.id!s}"
-        key_minute = f"{key_base:s}-minute"
-        key_hour = f"{key_base:s}-hour"
-
-        try:
-            count_minute = cache.incr(key_minute)
-        except ValueError:
-            cache.set(key_minute, 1, 60)
-            count_minute = 1
-
-        try:
-            count_hour = cache.incr(key_hour)
-        except ValueError:
-            cache.set(key_hour, 1, 3600)
-            count_hour = 1
-
-        if count_minute > 20 or count_hour > 150:
-            raise exceptions.Throttled()
-
         serializer = self.get_serializer(queryset, many=True)
         return response.Response(serializer.data)
 
@@ -184,10 +181,10 @@ class UserViewSet(mixins.UpdateModelMixin, viewsets.GenericViewSet):
     permission_classes = [permissions.IsSelf]
     queryset = models.User.objects.all().select_related("profile_contact")
     serializer_class = serializers.UserSerializer
+    throttle_classes = [BurstRateThrottle, SustainedRateThrottle]
 
     def list(self, request, *args, **kwargs):
         """Limit listed users by a query with throttle protection."""
-        user = self.request.user
         queryset = self.filter_queryset(self.get_queryset())
 
         # Exclude inactive contacts
@@ -208,26 +205,6 @@ class UserViewSet(mixins.UpdateModelMixin, viewsets.GenericViewSet):
                 )  # Lower value than in contacts viewset, to improve matching
                 .order_by("-similarity")
             )
-
-        # Throttle protection
-        key_base = f"throttle-user-list-{user.id!s}"
-        key_minute = f"{key_base:s}-minute"
-        key_hour = f"{key_base:s}-hour"
-
-        try:
-            count_minute = cache.incr(key_minute)
-        except ValueError:
-            cache.set(key_minute, 1, 60)
-            count_minute = 1
-
-        try:
-            count_hour = cache.incr(key_hour)
-        except ValueError:
-            cache.set(key_hour, 1, 3600)
-            count_hour = 1
-
-        if count_minute > 20 or count_hour > 150:
-            raise exceptions.Throttled()
 
         serializer = self.get_serializer(queryset, many=True)
         return response.Response(serializer.data)
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1,8 +1,8 @@
 """API endpoints"""
-from core import models
 from django.contrib.postgres.search import TrigramSimilarity
 from django.core.cache import cache
 from django.db.models import Func, OuterRef, Q, Subquery, Value
+
 from rest_framework import (
     decorators,
     exceptions,
@@ -11,6 +11,8 @@ from rest_framework import (
     response,
     viewsets,
 )
+
+from core import models
 
 from . import permissions, serializers
 
@@ -176,7 +178,12 @@ class UserViewSet(
     mixins.UpdateModelMixin,
     viewsets.GenericViewSet,
 ):
-    """User ViewSet"""
+    """
+    User ViewSet for all interactions with user infos and teams.
+
+    GET /api/users/&q=query
+        Return a list of all users whose email matches the query, completely or partially
+    """
 
     permission_classes = [permissions.IsSelf]
     queryset = models.User.objects.all().select_related("profile_contact")
@@ -206,13 +213,13 @@ class UserViewSet(
             queryset = (
                 queryset.annotate(similarity=similarity)
                 .filter(
-                    similarity__gte=0.05
-                )  # Value determined by testing (test_api_contacts.py)
+                    similarity__gte=0.01
+                )  # Value lowered to match every email containing the queries
                 .order_by("-similarity")
             )
 
         # Throttle protection
-        key_base = f"throttle-contact-list-{user.id!s}"
+        key_base = f"throttle-user-list-{user.id!s}"
         key_minute = f"{key_base:s}-minute"
         key_hour = f"{key_base:s}-hour"
 
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -16,6 +16,10 @@ from core import models
 
 from . import permissions, serializers
 
+EMAIL_SIMILARITY_THRESHOLD = 0.01
+# TrigramSimilarity threshold is lower for searching email than for names,
+# to improve matching results
+
 
 class NestedGenericViewSet(viewsets.GenericViewSet):
     """
@@ -200,9 +204,7 @@ class UserViewSet(mixins.UpdateModelMixin, viewsets.GenericViewSet):
             )
             queryset = (
                 queryset.annotate(similarity=similarity)
-                .filter(
-                    similarity__gte=0.01
-                )  # Lower value than in contacts viewset, to improve matching
+                .filter(similarity__gte=EMAIL_SIMILARITY_THRESHOLD)
                 .order_by("-similarity")
             )
 
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1,14 +1,8 @@
 """API endpoints"""
+from core import models
 from django.contrib.postgres.search import TrigramSimilarity
 from django.core.cache import cache
-from django.db.models import (
-    Func,
-    OuterRef,
-    Q,
-    Subquery,
-    Value,
-)
-
+from django.db.models import Func, OuterRef, Q, Subquery, Value
 from rest_framework import (
     decorators,
     exceptions,
@@ -17,8 +11,6 @@ from rest_framework import (
     response,
     viewsets,
 )
-
-from core import models
 
 from . import permissions, serializers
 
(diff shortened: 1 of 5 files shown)


src/backend/core/tests/test_api_users.py
```diff
@@ -17,8 +16,11 @@ def test_api_users_list_anonymous():
     factories.UserFactory()
     client = APIClient()
     response = client.get("/api/v1.0/users/")
-    assert response.status_code == 404
-    assert "Not Found" in response.content.decode("utf-8")
+    assert response.status_code == 401
+    print(response.json())
+    assert "Authentication credentials were not provided." in response.content.decode(
+        "utf-8"
+    )
```
mjeammet, line 23: Adding the `list`method to the endpoint changed several tests. I'll comment where doubts remain.

src/backend/core/tests/test_api_users.py
```diff
@@ -36,6 +38,142 @@ def test_api_users_list_authenticated():
...
+    assert user_ids == [str(dave.id)]
+
+    # Unaccented partial query
+    response = APIClient().get(
+        "/api/v1.0/users/?q=david", HTTP_AUTHORIZATION=f"Bearer {jwt_token}"
+    )
+
+    assert response.status_code == 200
+    user_ids = [user["id"] for user in response.json()]
+    assert user_ids == [str(dave.id)]
```
mjeammet, line 118: I'm unsure if that's relevant : we already ascertained partial query worked in a previous test.

src/backend/core/tests/test_api_users.py
```diff
@@ -36,6 +38,142 @@ def test_api_users_list_authenticated():
...
+
+    # Unaccented partial name
+    response = APIClient().get(
+        "/api/v1.0/users/?q=DAVID", HTTP_AUTHORIZATION=f"Bearer {jwt_token}"
+    )
+
+    assert response.status_code == 200
+    user_ids = [user["id"] for user in response.json()]
+    assert user_ids == [str(dave.id)]
```
mjeammet, line 146: same comment as `test_api_users_authenticated_list_uppercase_content`

src/backend/core/tests/test_api_users.py
```diff
@@ -17,8 +16,11 @@ def test_api_users_list_anonymous():
...
-    assert response.status_code == 404
-    assert "Not Found" in response.content.decode("utf-8")
+    assert response.status_code == 401
+    print(response.json())
+    assert "Authentication credentials were not provided." in response.content.decode(
+        "utf-8"
+    )
 
 
 def test_api_users_list_authenticated():
```
mjeammet, line 26: This test is now broken because we succesfully retrieve a list (200 != 404). I feel that listing all users should be allowed, as query searches can list all users anyway. Do you confirm, @sampaccoud ?

src/backend/core/api/viewsets.py
```diff
@@ -184,12 +176,69 @@ class UserViewSet(
...
+            is_active=True,
+        )
+
+        # Search by case-insensitive and accent-insensitive trigram similarity
+        if query := self.request.GET.get("q", ""):
+            query = Func(Value(query), function="unaccent")
+            similarity = TrigramSimilarity(
+                Func("email", function="unaccent"),
+                query,
+            )
```
mjeammet, line 210: Shouldn't we allow searching by `full_name`, `short_name` too, as for contacts ?

src/backend/core/tests/test_api_users.py
```diff
@@ -32,8 +34,144 @@ def test_api_users_list_authenticated():
...
+    # "Nicole Foole" matches twice on "ole"
+    assert user_ids == [str(nicole.id), str(frank.id)]
+
+    response = APIClient().get(
+        "/api/v1.0/users/?q=ool", HTTP_AUTHORIZATION=f"Bearer {jwt_token}"
+    )
+
+    assert response.status_code == 200
+    user_ids = [user["id"] for user in response.json()]
+    assert user_ids == [str(nicole.id), str(frank.id)]
```
mjeammet, line 90: This test always fails and I can't seem to figure out why :/ 
response is always [ ]. 

I commented throttle protection to test it out but it kept on failing.

src/backend/core/tests/test_api_users.py
```diff
@@ -32,8 +34,144 @@ def test_api_users_list_authenticated():
...
+    assert response.status_code == 200
+    assert len(response.json()) == 3
+
+
+def test_api_users_authenticated_list_by_email():
+    """
+    Authenticated users should be able to search users with a case insensitive and
+    partial query on the email.
+    """
+    identity = factories.IdentityFactory(email="tester@ministry.fr")
```
mjeammet, line 46: I'm setting the email for the authenticated user because tests sometimes fail because of random emails containing queried strings.

src/backend/core/tests/test_api_users.py
mjeammet: It was because of the similarity threshold defined at viewsets.py:214. `0.05` filtered out this two emails as irrelevant, while `0.01` succesfully returns expected result. But other tests fail. 

As this value of '0.05' seem to have been benchmarked, I'm setting the expected result of this current test as `[]`.

src/backend/core/api/viewsets.py
```diff
@@ -184,12 +178,69 @@ class UserViewSet(
     mixins.UpdateModelMixin,
     viewsets.GenericViewSet,
 ):
-    """User ViewSet"""
+    """
+    User ViewSet for all interactions with user infos and teams.
+
+    GET /api/users/&q=query
+        Return a list of all users whose email matches the query, completely or partially
```
lebaudantoine, line 185: nit: the docstring could be a bit more descriptive, or explicit about the purpose of the view.

src/backend/core/api/viewsets.py
```diff
@@ -184,12 +178,69 @@ class UserViewSet(
...
+        try:
+            count_hour = cache.incr(key_hour)
+        except ValueError:
+            cache.set(key_hour, 1, 3600)
+            count_hour = 1
+
+        if count_minute > 20 or count_hour > 150:
+            raise exceptions.Throttled()
+
+        serializer = self.get_serializer(queryset, many=True)
```
lebaudantoine, line 241: Values as `0.01``20` (etc) feel a bit as magic values. Could we consider extracting them to constant with a descriptive and explicit naming? Is it overkilled?

src/backend/core/api/viewsets.py
```diff
@@ -184,12 +178,69 @@ class UserViewSet(
...
+        queryset = self.filter_queryset(self.get_queryset())
+
+        if not user.is_authenticated:
+            return queryset.none()
+
+        # Exclude contacts that:
+        queryset = queryset.filter(
+            # - inactive users
+            is_active=True,
+        )
```
lebaudantoine, line 204:
```suggestion
        # Exclude inactive contacts
        queryset = queryset.filter(
            is_active=True,
        )
```

src/backend/core/api/viewsets.py
```diff
@@ -184,12 +178,69 @@ class UserViewSet(
...
+
+        # Search by case-insensitive and accent-insensitive trigram similarity
+        if query := self.request.GET.get("q", ""):
+            query = Func(Value(query), function="unaccent")
+            similarity = TrigramSimilarity(
+                Func("email", function="unaccent"),
+                query,
+            )
```
lebaudantoine, line 212:
```suggestion
        if query := self.request.GET.get("q", ""):
            similarity = TrigramSimilarity(
                Func("email", function="unaccent"),
                Func(Value(query), function="unaccent"),
            )
```

src/backend/core/api/viewsets.py
```diff
@@ -184,12 +178,69 @@ class UserViewSet(
...
+            queryset = (
+                queryset.annotate(similarity=similarity)
+                .filter(
+                    similarity__gte=0.01
+                )  # Value lowered to match every email containing the queries
```
lebaudantoine, line 217: I am not sure to fully understand this comment, let's discuss it

src/backend/core/api/viewsets.py
```diff
@@ -184,12 +178,69 @@ class UserViewSet(
...
+            count_minute = 1
+
+        try:
+            count_hour = cache.incr(key_hour)
+        except ValueError:
+            cache.set(key_hour, 1, 3600)
+            count_hour = 1
+
+        if count_minute > 20 or count_hour > 150:
+            raise exceptions.Throttled()
```
lebaudantoine, line 239: Could we extract this logic part in a dedicated function? It will be reusable and easier to test.

lebaudantoine (commented):
Thx for your work, so coool 😊
I have few suggestions on the endpoint itself, I would like to discuss first, before reviewing your tests. 
I will continue my review as soon as we resolve these first conversations.
Cheers

src/backend/core/api/viewsets.py
mjeammet: Willingly. After our daily ?

src/backend/core/tests/test_api_users.py
```diff
@@ -32,8 +34,145 @@ def test_api_users_list_authenticated():
...
+
+def test_api_contacts_list_authenticated_accented_query():
+    """Accented content should be found by unaccented query."""
+    user = factories.UserFactory(email="tester@ministry.fr")
+    factories.IdentityFactory(user=user, email=user.email)
+    jwt_token = OIDCToken.for_user(user)
+
+    helene = factories.UserFactory(email="helene.bowman@work.com")
+
+    # Unaccented full email
```
AntoLC, line 158:
```suggestion
    # Accented full email
```

src/backend/core/tests/test_api_users.py
```diff
@@ -32,8 +34,145 @@ def test_api_users_list_authenticated():
...
+    response = APIClient().get(
+        "/api/v1.0/users/?q=DAVID.BOWMAN@WORK.COM",
+        HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
+    )
+
+    assert response.status_code == 200
+    user_ids = [user["id"] for user in response.json()]
+    assert user_ids == [str(dave.id)]
+
+    # Unaccented partial email
```
AntoLC, line 140:
```suggestion
    # Partial uppercase email
```

src/backend/core/tests/test_api_users.py
```diff
@@ -32,8 +34,145 @@ def test_api_users_list_authenticated():
...
+
+def test_api_users_list_authenticated_capital_query():
+    """Upper case query should find lower case content."""
+    user = factories.UserFactory(email="tester@ministry.fr")
+    factories.IdentityFactory(user=user, email=user.email)
+    jwt_token = OIDCToken.for_user(user)
+
+    dave = factories.UserFactory(email="david.bowman@work.com")
+
+    # Unaccented full query
```
AntoLC, line 130:
```suggestion
    # Full uppercase query
```

src/backend/core/tests/test_api_users.py
```diff
@@ -32,8 +34,145 @@ def test_api_users_list_authenticated():
...
+    response = APIClient().get(
+        "/api/v1.0/users/?q=david.bowman@intensework.com",
+        HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
+    )
+
+    assert response.status_code == 200
+    user_ids = [user["id"] for user in response.json()]
+    assert user_ids == [str(dave.id)]
+
+    # Unaccented partial query
```
AntoLC, line 112:
```suggestion
    # Partial query
```

src/backend/core/tests/test_api_users.py
AntoLC: It is funny this similarity variable, if I have 3 emails [toto, poto, o]:
- Search: `o` == [o]
- Search: `to` == [toto, poto]
- Search: `ot` == [o]

src/backend/core/api/viewsets.py
```diff
@@ -184,12 +178,69 @@ class UserViewSet(
...
+            return queryset.none()
+
+        # Exclude contacts that:
+        queryset = queryset.filter(
+            # - inactive users
+            is_active=True,
+        )
+
+        # Search by case-insensitive and accent-insensitive trigram similarity
+        if query := self.request.GET.get("q", ""):
```
AntoLC, line 207: I have a question, does is have to be `q` ?  Why not email ?

src/backend/core/api/viewsets.py
AntoLC, line 241: Could be nice to add the email in the serializer? 
We have that now:
```json
{
        "id": "01c9c97f-32e1-4991-ae96-7faaec61c447",
        "data": {},
        "language": "en-us",
        "timezone": "UTC",
        "is_device": false,
        "is_staff": false
}
```

src/backend/core/api/viewsets.py
AntoLC: Could we implement that maybe ? Seems more the DRH way.
https://www.django-rest-framework.org/api-guide/throttling/

src/backend/core/tests/test_api_users.py
```diff
@@ -322,7 +463,7 @@ def test_api_users_delete_list_anonymous():
     client = APIClient()
     response = client.delete("/api/v1.0/users/")
 
-    assert response.status_code == 404
+    assert response.status_code == 401
```
AntoLC, line 466: Question, how did these tests were passing before this PR ?

AntoLC (commented):
I invited myself 😅 

I have the feeling that the second commit is a fix of the first one, plus some tests, could they be squashed together.

A few comment / question, but seems a nice research feature!

src/backend/core/tests/test_api_users.py
AntoLC: Ok, I found the answer in your previous comment, sorry.

src/backend/core/tests/test_api_users.py
AntoLC: IMO you can let it.

src/backend/core/api/viewsets.py
AntoLC: I would need them though:
image

src/backend/core/api/viewsets.py
mjeammet: isn't the purpose of the view to list users ? I'll add a little explanation as to the way we match the email.

src/backend/core/api/viewsets.py
```diff
@@ -184,12 +178,69 @@ class UserViewSet(
...
     queryset = models.User.objects.all().select_related("profile_contact")
     serializer_class = serializers.UserSerializer
 
+    def list(self, request, *args, **kwargs):
+        """Limit listed users by a query with throttle protection."""
+        user = self.request.user
+        queryset = self.filter_queryset(self.get_queryset())
+
+        if not user.is_authenticated:
+            return queryset.none()
```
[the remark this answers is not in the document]
mjeammet: We could add the "login_required" decorator on the methods, or find a mixin for the whole viewset !

src/backend/core/api/viewsets.py
mjeammet: For now we only filter by email but I imagine we'll later use that same field to search for names or organisation.

src/backend/core/api/viewsets.py
mjeammet: Hm, it feels like we'll have to store them finally. We'll have to discuss that point with Samuel

src/backend/core/api/viewsets.py
mjeammet: "lowered" par rapport à quoi.

src/backend/core/tests/test_api_users.py
mjeammet, line 158:
```suggestion
    # Accented full query
```
Only the query can be accented, as django won't let you add accentuated characters to an email field.

mjeammet:
> I invited myself 😅
> 
> I have the feeling that the second commit is a fix of the first one, plus some tests, could they be squashed together.
> 
> A few comment / question, but seems a nice research feature!

Welcome ! Next time I'll invite you directly but feel free to jump in anytime. 

I'll squash everything in the end, especially because i'll add a few small commits to include your comments.

src/backend/core/api/viewsets.py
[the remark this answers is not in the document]
mjeammet: I changed the global setting to "isAuthenticated". Now, by default all endpoints require an authenticated user, and we can change that behaviour ponctually if needed.

WDYT ? I thought we could start simple and add a permission file later, if needed.

src/backend/core/tests/test_api_users.py
mjeammet: OK, thanks !

src/backend/core/api/viewsets.py
mjeammet: I don't know what Samuel had in mind but I feel like this could be sumed up in settings, as suggested in Anto's link : 

```python
REST_FRAMEWORK = {
    'DEFAULT_THROTTLE_CLASSES': [
        'rest_framework.throttling.UserRateThrottle'
    ],
    'DEFAULT_THROTTLE_RATES': {
        'user': '150/hour',
        'user': '20/minute'
    }
}
```

mjeammet: no Anon here because Anon is capped at 0/second.

src/backend/core/api/viewsets.py
mjeammet: I added it to the serializer and I slightyly modified a test to it to pass. 

Two tests remain broken if I don't list `email` as a read_only fied, and I don't get why. I thus added it to the read_only fields, as I didn't want to spend too much time on it, not being the focus of this PR.

src/backend/core/api/viewsets.py
```diff
@@ -180,22 +172,50 @@ def perform_create(self, serializer):
...
+    """
 
     permission_classes = [permissions.IsSelf]
     queryset = models.User.objects.all().select_related("profile_contact")
     serializer_class = serializers.UserSerializer
+    throttle_classes = [BurstRateThrottle, SustainedRateThrottle]
+
+    def list(self, request, *args, **kwargs):
+        """Limit listed users by a query with throttle protection."""
+        queryset = self.filter_queryset(self.get_queryset())
```
mjeammet, line 192: I don't see what this method does, @sampaccoud

src/backend/core/api/viewsets.py
```diff
@@ -11,6 +10,7 @@
     response,
     viewsets,
 )
+from rest_framework.throttling import UserRateThrottle
```
AntoLC, line 13: I think we should write a test to asset that it is working, plus to be secure in the future that it is still working when we upgrade something, I wrote a test for you, the test is working, but you can probably make it cleaner:

`./src/backend/core/tests/test_throttle.py`
```py
"""
Test Throttle in People's app.
"""

from .utils import OIDCToken

from django.test.utils import override_settings

import pytest
from rest_framework.test import APIClient

from core import factories, models
from core.api import serializers

from .utils import OIDCToken

pytestmark = pytest.mark.django_db


@override_settings(REST_FRAMEWORK={
  'DEFAULT_THROTTLE_RATES': {
    'burst': '2/min',
    'sustained': '2/min'
}})
def test_throttle():
    identity = factories.IdentityFactory()
    user = identity.user
    contact = factories.ContactFactory(owner=user)
    user.profile_contact = contact
    user.save()
    jwt_token = OIDCToken.for_user(user)

    # Let's have 5 contacts in database:
    assert user.profile_contact is not None  # Excluded because profile contact
    base_contact = factories.BaseContactFactory()  # Excluded because overriden
    factories.ContactFactory(
        base=base_contact
    )  # Excluded because belongs to other user

    client = APIClient()
    # some end point you want to test (in this case it's a public enpoint that doesn't require authentication
    _url = '/api/v1.0/contacts/'
    
    # loop to activate the protection
    for i in range(0, 5):
        client.get(_url, HTTP_AUTHORIZATION=f"Bearer {jwt_token}")

    # this call should err
    response = client.get(_url, HTTP_AUTHORIZATION=f"Bearer {jwt_token}")
    # 429 - too many requests
    assert response.status_code == 429
```

AntoLC (commented):
Nice job!
A comment about the test on the throttling, and I think we are good on my side.

About the commits I find they are mostly correct, IMO you don't have to squash them except the 3 first that can be only one, I can give you some support to squash them if you want to go in this direction:
[✨(api) search users by email](https://github.com/numerique-gouv/people/pull/16/commits/b5ef8ae2dfdb9ddb3d763e8264228e837917b68a)
[✅(api) test list users by email](https://github.com/numerique-gouv/people/pull/16/commits/397e4abe1de1a89c501a27f9c0d4c38ce104670a)
[💡(api) improve documentation and test comments](https://github.com/numerique-gouv/people/pull/16/commits/d792b11e12bfdde4ef895c75833889ab9a402d9d)

src/backend/core/api/viewsets.py
mjeammet: I made a very basic version of this test 
- instead of contacts, I use /users/ because I don't have to create anything
- I'm having a hard time overrind the throttle burst settings, despite using [related pytest doc](https://pytest-django.readthedocs.io/en/latest/configuring_django.html#overriding-individual-settings) and [django doc](https://docs.djangoproject.com/fr/4.2/topics/testing/tools/)

AntoLC: That doesn't override ? 
```py
@override_settings(REST_FRAMEWORK={
  'DEFAULT_THROTTLE_RATES': {
    'burst': '2/min',
    'sustained': '2/min'
}})
```
The test passed on my side, so it must override it.

AntoLC: Ok, indeed, keep what you have done. 
Like that it works, but it doesn't seem very clean:
```py
...
from rest_framework.decorators import api_view

@override_settings(REST_FRAMEWORK={
  'DEFAULT_THROTTLE_RATES': {
    'burst': '2/min',
    'sustained': '2/min'
}})
def test_throttle(request):
 ...
```

AntoLC (approved):
✅ 🚀 !

src/backend/core/api/viewsets.py
mjeammet: On my side it only works when making 20+1 calls, the standard protection defined in settings.py. Any amount under 21 and the test fails (`assert 200 != 429`)

mjeammet: Ok we investigated and, for the same code : 

- commands `make test` and `bin/pytest -k test_throttle` return this test failing, for assert 200 != 429
- command `make test-back core/tests/test_throttle.py` make it pass

src/backend/core/api/viewsets.py
```diff
@@ -181,21 +173,52 @@ def perform_create(self, serializer):
...
+            if query := self.request.GET.get("q", ""):
+                similarity = TrigramSimilarity(
+                    Func("email", function="unaccent"),
+                    Func(Value(query), function="unaccent"),
+                )
+                queryset = (
+                    queryset.annotate(similarity=similarity)
+                    .filter(similarity__gte=EMAIL_SIMILARITY_THRESHOLD)
+                    .order_by("-similarity")
+                )
```
lebaudantoine, line 213: nit: we could early return, this [video](https://www.youtube.com/watch?v=CFRhGnuXG-4) is amazing at explaining why

src/backend/core/tests/test_throttle.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    identity = factories.IdentityFactory()
+    user = identity.user
+    jwt_token = OIDCToken.for_user(user)
+
+    client = APIClient()
+    endpoint = "/api/v1.0/users/"
+
+    # loop to activate the protection
+    for _ in range(0, 20):
```
lebaudantoine, line 26: nit: add a comment designating the setting being tested.

src/backend/core/tests/test_api_users.py
```diff
@@ -17,13 +21,15 @@ def test_api_users_list_anonymous():
...
+    assert "Authentication credentials were not provided." in response.content.decode(
+        "utf-8"
+    )
 
 
 def test_api_users_list_authenticated():
     """
-    Authenticated users should not be able to list users.
+    Authenticated users should be able to list all users.
```
lebaudantoine, line 32: let's discuss it orally, and see if this is an issue.

src/backend/core/tests/test_api_users.py
```diff
@@ -32,8 +38,184 @@ def test_api_users_list_authenticated():
...
     )
-    assert response.status_code == 404
-    assert "Not Found" in response.content.decode("utf-8")
+    assert response.status_code == 200
+    assert len(response.json()["results"]) == 3
+
+
+def test_api_users_authenticated_list_by_email():
+    """
+    Authenticated users should be able to search users with a case insensitive and
```
lebaudantoine, line 47: nit:

```suggestion
    Authenticated users should be able to search users with a case-insensitive and
```

src/backend/core/tests/test_api_users.py
```diff
@@ -32,8 +38,184 @@ def test_api_users_list_authenticated():
...
+
+def test_api_users_authenticated_list_uppercase_content():
+    """Upper case content should be found by lower case query."""
+    user = factories.UserFactory(email="tester@ministry.fr")
+    factories.IdentityFactory(user=user, email=user.email)
+    jwt_token = OIDCToken.for_user(user)
+
+    dave = factories.UserFactory(email="DAVID.BOWMAN@INTENSEWORK.COM")
+
+    # Unaccented full adress
```
lebaudantoine, line 106: nit:
```suggestion
    # Unaccented full address
```

lebaudantoine (approved):
Nit-picking ! LGTM

src/backend/core/api/viewsets.py
mjeammet: OK looks interesting, I'm saving it for later.

PR #24, opened 2024-01-29
👷(ci) sops: Add age key

rouja:
Add key for Antoine Lebaud


lebaudantoine (approved):
LGTM, tested locally ✅

PR #36, opened 2024-02-01
✨(models) add slug field to Team model

mjeammet:
## Purpose

Add unique slugs for team objects, to avoid duplicated team names when looking for a team.

## Proposal

Description...

- [x] add field and migrations file
- [x] write tests for expected, unaccented, no-unicode, lower-case slugs
- [x] write tests making sure no slug can be duplicated (neither at creation nor update)
- [x] fix test suite


src/backend/core/tests/teams/test_core_api_teams_create.py
```diff
@@ -48,3 +48,35 @@ def test_api_teams_create_authenticated():
...
+    response = APIClient().get(
+        "/api/v1.0/teams/my-team/",
+        {
+            "name": "my team",
+        },
+        HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
+    )
+
+    print(response.json())
+    assert response.status_code == 200
```
lebaudantoine, line 82: EOF file, should we add a linter rule?

src/backend/core/tests/teams/test_core_api_teams_create.py
```diff
@@ -48,3 +48,35 @@ def test_api_teams_create_authenticated():
...
+
+    response = APIClient().get(
+        "/api/v1.0/teams/my-team/",
+        {
+            "name": "my team",
+        },
+        HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
+    )
+
+    print(response.json())
```
lebaudantoine, line 81: Wrongly committed print statement 🤪

src/backend/core/api/serializers.py
```diff
@@ -131,15 +133,20 @@ class TeamSerializer(serializers.ModelSerializer):
...
 
     def get_abilities(self, team) -> dict:
         """Return abilities of the logged-in user on the instance."""
         request = self.context.get("request")
         if request:
             return team.get_abilities(request.user)
         return {}
+
+    def get_slug(self, instance):
+        return slugify(instance.name)
```
lebaudantoine, line 152: EOF

src/backend/core/api/serializers.py
```diff
@@ -131,15 +133,20 @@ class TeamSerializer(serializers.ModelSerializer):
...
     accesses = TeamAccessSerializer(many=True, read_only=True)
+    slug = serializers.SerializerMethodField()
 
     class Meta:
         model = models.Team
-        fields = ["id", "name", "accesses", "abilities"]
-        read_only_fields = ["id", "accesses", "abilities"]
+        fields = ["id", "name", "accesses", "abilities", "slug"]
+        read_only_fields = ["id", "accesses", "abilities", "slug"]
+        lookup_field = 'slug'
```
lebaudantoine, line 142: did you lint your code? mixing `"` and `'` feels weird.

src/backend/core/models.py
```diff
@@ -358,6 +360,11 @@ def get_abilities(self, user):
             "manage_accesses": is_owner_or_admin,
         }
 
+    def save(self, *args, **kwargs):
```
lebaudantoine, line 363: nit: always good to have docstring, linter should fail

src/backend/core/models.py
```diff
@@ -315,6 +316,7 @@ class Team(BaseModel):
     """
 
     name = models.CharField(max_length=100)
+    slug = models.SlugField(max_length=100, unique=True, null=False)
```
lebaudantoine, line 319: Corner case: (which won't happen often)
Should they have equal max-length? 
does some characters slugify to a longer version?
ex: is `&` slugified to `and` ?

src/backend/core/tests/teams/test_core_api_teams_create.py
```diff
@@ -48,3 +48,35 @@ def test_api_teams_create_authenticated():
...
+
+    response = APIClient().post(
+        "/api/v1.0/teams/",
+        {
+            "name": "my team",
+        },
+        HTTP_AUTHORIZATION=f"Bearer {jwt_token}",
+    )
+
+    assert response.status_code == 201
```
lebaudantoine, line 68: use the constant HTTP status code?

src/backend/core/tests/teams/test_core_api_teams_create.py
```diff
@@ -48,3 +48,35 @@ def test_api_teams_create_authenticated():
...
+    """
+    Creating teams should automatically generate their slug.
+    """
+    identity = IdentityFactory()
+    jwt_token = OIDCToken.for_user(identity.user)
+
+    response = APIClient().post(
+        "/api/v1.0/teams/",
+        {
+            "name": "my team",
```
lebaudantoine, line 63: Nice one ! I would parametrize this tests to cover more case, as strings containing special characters, double spaces, etc..

src/backend/core/api/serializers.py
lebaudantoine: Missing docstring !

src/backend/core/admin.py
```diff
@@ -108,4 +108,5 @@ class TeamAdmin(admin.ModelAdmin):
         "created_at",
         "updated_at",
     )
+    prepopulated_fields = {"slug": ("name",)}
```
lebaudantoine, line 111: Should we prevent editing this field, as initially we introduced a slug, to make sure all teams' name would be unique?

src/backend/core/api/viewsets.py
```diff
@@ -243,6 +243,7 @@ class TeamViewSet(
     permission_classes = [permissions.AccessPermission]
     serializer_class = serializers.TeamSerializer
     queryset = models.Team.objects.all()
+    lookup_field = 'slug'
```
lebaudantoine, line 246: nit: double quote

lebaudantoine (commented):
Nice one

src/backend/core/tests/teams/test_core_api_teams_create.py
mjeammet: Oups, I must have forgotten to run "make lint". Should we add it to precommits scripts ?

lebaudantoine: We should create a proper pre-commit hook, I take it, will open a PR !

src/backend/core/tests/teams/test_core_api_teams_create.py
mjeammet: There is one. My linter is now up and running aaagaïn !

src/backend/core/api/viewsets.py
mjeammet: I removed this lookup field as we'll keep on using ids for URIs

src/backend/core/models.py
mjeammet: Good question ! 

I tested it, to be sure. It seems that `&` is interpreted as a special caracter, and thus removed because I used `validate_slug` and not `validate_unicode_slug` ([see django docs](https://docs.djangoproject.com/en/4.2/ref/validators/#django.core.validators.validate_slug)).

I don't see another scenario where len(slug) > len(name).

src/backend/core/tests/teams/test_core_api_teams_create.py
mjeammet: Done !

src/backend/core/tests/teams/test_core_api_teams_create.py
mjeammet: Great idea ! 

Should I use param with tuples, such as [("my team", "my team"), ("my     team", "my team")] to make sure slugify is the expected str ?

mjeammet: OK I did that and it works wonderfully. I love it, thanks for reminding me that existed ! 

I used the parameters on another test that the one you were pointing out : 
- `test_api_teams_create_authenticated_slugify_name` tests a slug is added 
- `test_api_teams_create_authenticated_expected_slug` tests all the aforementioned scenarios (previously unaccented scenario only)

src/backend/core/tests/teams/test_core_api_teams_create.py
```diff
@@ -2,6 +2,11 @@
 Tests for Teams API endpoint in People's core app: create
 """
 import pytest
+from rest_framework.status import (
```
sampaccoud, line 5: Imo it's better to use raw values in tests:
- tests should be as simple and understandable as possible
- If the value of HTTP_401_AUORIZED changes one day, I would like my tests to break and decide what to do... 

ok, the change in value may never happen here... but I assume that what drives this change is that we would want to replace all other raw values by constants in our tests and I think it's an anti-pattern.

src/backend/core/tests/teams/test_core_api_teams_create.py
```diff
@@ -74,33 +72,42 @@ def test_api_teams_create_authenticated_slugify_name():
     assert team.slug == "my-team"
 
 
-def test_api_teams_create_authenticated_unaccented_slugify_name():
+@pytest.mark.parametrize(
```
sampaccoud, line 75: :medal_sports:

src/backend/core/migrations/0002_team_slug.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 5.0.1 on 2024-02-01 14:24
```
sampaccoud, line 1: Until we go to production, I propose to modify 0001_initial instead of adding migrations.
adding migrations when we still break things a lot would mean writing data migrations which we don't want to do until we have to! 
maybe add a warning in the README that this project is not yet for production and that we will still break things like the database schema?

src/backend/core/tests/test_models_teams.py
```diff
@@ -46,6 +46,12 @@ def test_models_teams_name_max_length():
         factories.TeamFactory(name="a " * 51)
 
 
+def test_models_teams_slug_empty():
+    """Slug field should not be empty."""
+    with pytest.raises(ValidationError, match="This field cannot be blank."):
+        models.Team.objects.create(slug="")
+
```
sampaccoud, line 53: you should test your save method override:
- the slug is automatically computed
- a slug that is already generated will not change if we change the name of the team and save it (this is what we want right?)

src/backend/demo/management/commands/create_demo.py
```diff
@@ -143,6 +143,8 @@ def create_demo(stdout):
             queue.push(
                 models.Team(
                     name=f"Team {i:d}",
+                    # slug should be automatic but bulk_create doesn't use save
+                    slug=f"Team-{i:d}",
```
sampaccoud, line 147: while you're at it, make it a real slug :stuck_out_tongue: 
```suggestion
                    slug=f"team-{i:d}",
```

src/backend/core/tests/teams/test_core_api_teams_update.py
```diff
@@ -121,7 +125,8 @@ def test_api_teams_update_authenticated_administrators():
 
 
 def test_api_teams_update_authenticated_owners():
-    """Administrators of a team should be allowed to update it."""
+    """Administrators of a team should be allowed to update it,
+    appart from read-only fields."""
```
sampaccoud, line 129:
```suggestion
    apart from read-only fields."""
```

src/backend/core/tests/teams/test_core_api_teams_retrieve.py
```diff
@@ -1,16 +1,12 @@
 """
 Tests for Teams API endpoint in People's core app: retrieve
 """
-import random
-from collections import Counter
-from unittest import mock
-
```
sampaccoud, line 7: do we have a linter configuration problem? This should have been raised by pylint right?

src/backend/core/api/serializers.py
```diff
@@ -131,15 +133,20 @@ class TeamSerializer(serializers.ModelSerializer):
 
     abilities = serializers.SerializerMethodField(read_only=True)
     accesses = TeamAccessSerializer(many=True, read_only=True)
+    slug = serializers.SerializerMethodField()
```
sampaccoud, line 136: actually this is not required. You should use the `slug` field from the model directly...

src/backend/core/api/serializers.py
```diff
@@ -131,15 +133,20 @@ class TeamSerializer(serializers.ModelSerializer):
...
     def get_abilities(self, team) -> dict:
         """Return abilities of the logged-in user on the instance."""
         request = self.context.get("request")
         if request:
             return team.get_abilities(request.user)
         return {}
+
+    def get_slug(self, instance):
+        """Return slug from the team's name."""
+        return slugify(instance.name)
```
sampaccoud, line 152: ditto
```suggestion
```

src/backend/core/tests/test_models_teams.py
lebaudantoine: I am not sure that what we want. Initially, we introduce the slug, to avoid having duplicated Team's name. Having a slug that doesn't reflect the current team's name break this check.

I agree that in general, slug doesn't change after their initial creation ... What should we do?

AntoLC (commented):
2 things:
- In my tests, during the creation (POST `/api/v1.0/teams/`) the unicity on slug is correctly indicated, but if I update the name of my team with the same name as another team (PUT `/api/v1.0/teams/{id}/`), the GET will return on both teams the same slug, yet the slug in the table are different.
- the slug is not updated when I update the name of my team.

src/backend/core/tests/test_models_teams.py
mjeammet: To me, the slug is an unique version of the name. When the name changes, its slug should change accordingly. Ids are still unchangeable, if we want a secure way to access the team.

src/backend/core/tests/teams/test_core_api_teams_create.py
mjeammet: thanks !

src/backend/core/migrations/0002_team_slug.py
mjeammet: Gotcha, I'll modify 0001_initial and add a note on README.

src/backend/core/api/serializers.py
mjeammet: Well, it is indeed useless if you only use the model and its modified `save`. In cases where you use the serializer directly though, the slug field comes empty without a `get_slug` method. And I need to define "slug" here to do that.

src/backend/core/api/serializers.py
mjeammet: see answer on slug field, l.136.

AntoLC (approved):
✅  Works as expected on my side

Take care of your commits ;)

src/backend/core/models.py
```diff
@@ -332,6 +334,11 @@ class Meta:
     def __str__(self):
         return self.name
 
+    def save(self, *args, **kwargs):
+        """Overriding save function to add a slug whenever there is none."""
```
sampaccoud, line 338:
```suggestion
        """Overriding save function to compute the slug."""
```

src/backend/core/tests/teams/test_core_api_teams_create.py
sampaccoud: except in tests yes. Not going to bikeshed about it though.

src/backend/core/tests/teams/test_core_api_teams_retrieve.py
sampaccoud: but the CI linter should have failed... the lint-back job in the CI doesn't seem to cover tests :thinking:

src/backend/core/api/serializers.py
sampaccoud: Ok in this case I propose to call a common method on the model to have unicity of the slug computation:
```
instance.compute_slug
```
and on the model:
```
def get_slug(self):
    return self.slug or slugify(self.name)
```

src/backend/core/tests/test_models_teams.py
sampaccoud: Agree with changing the slug each time the name changes! Could be tested :wink: 
```
def test_models_teams_slug_updated_with_name():
    """The slug field should be updated when the name changes."""
    team = factories.TeamFactory()

    team.name = "new name"
    team.save()
    team.refresh_from_db()

    assert team.slug == "new-name"
```

src/backend/core/tests/teams/test_core_api_teams_retrieve.py
mjeammet: From what I see on that other PR, it seems to be working as expected : https://github.com/numerique-gouv/people/actions/runs/7818569220/job/21329100153?pr=44

sampaccoud (approved):

Issue #25, opened 2024-01-29
User invitation


lebaudantoine:
## Feature Request

**Describe the solution you'd like**

Allow user invitation and enhance group management by implementing a streamlined process:

1. User Invitation:
- Create an endpoint for inviting users to Desk based on provided email addresses.
- Generate pending identities for invited users.

2. First-time Connection:
- Check for pending identities when a user connects for the first time.
- Update/Link user identity if a pending identity exists.

3. Efficient Management:
- Allow adding pending identities to groups without re-entering email.
- Enable search among pending identities and those linked to active users.

mjeammet:
Sounds great ! A few extra questions : 

1) What's the model behind pending identity ? It could be an Identity with an `is_active` field, but that would mean creating Identities with unknown/blank/null `sub` field, which messes with it's unicity/mandatory constraint. Plus you would have to update the `sub` upon first connexion. 

Also, in addition to the invited email, we need to save pending group infos (group, role, date of addition, etc). Should we use the `TeamAccess` model, with an inactive User ? 

2) If you search for one of my adresses that I don't use with the SSO, it would be nice to be suggested an existing identity. I wonder if it wouldn't already be the case with the TrigramSimilarity research and email adresses almost always following the same pattern. I wonder if there's also a GDPR-related legal concern to have here.

lebaudantoine:
Brainsto:

1.
Might be cool to introduce an invitation model?
It would include:
- FK to the sender (User)
- FK to the group 
- Email of the recipient, that would be checked when new user log-in.

2. I guess the Trigram Similarity would show emails that contain your first name and last name. The current search feature would need to be improved in the longer run, agree with you

sampaccoud:
+1 for a separate invitation model. Separation of concerns?

sampaccoud:
> Generate pending identities for invited users.
This does not need to be done. You invite the users to join by sending them an email. When they click in the email, they must start by logging in to Desk... this is when their identity is automatically created. 
We then see that there is a pending invitation for this email address and can add the user to the group as indicated in the invitation... Don't forget to delete or archive the invitation.

mjeammet:
I don't quite remember what we said the other day during the standup : there should be a CRUD for this model, right ? Numerous uses : 
Create ✅ upon creation
Read ✅ upon listing group members, I want to see pending invitations
Read details ❓ 
Update ❓ to update role after first invite ?
Delete ✅ if I want to cancel an invite

PR #44, opened 2024-02-07
🗃️(model) create invitation model

mjeammet:
## ❓ Purpose

Create `Invitation` model, factory and related tests, in order to facilitate incoming developments surrounding user invitation.

We chose to create a separate dedicated model for separation of concerns, see https://github.com/numerique-gouv/people/issues/25

## 💌 Proposal

- [x] Write model
- [x] Add factory to ease test creation
- [x] Write basic unit tests for each field

### DIFF
--- a/src/backend/core/admin.py
+++ b/src/backend/core/admin.py
@@ -119,3 +119,31 @@ class TeamAdmin(admin.ModelAdmin):
         "updated_at",
     )
     search_fields = ("name",)
+
+
+@admin.register(models.Invitation)
+class InvitationAdmin(admin.ModelAdmin):
+    """Admin interface to handle invitations."""
+
+    fields = (
+        "email",
+        "team",
+        "role",
+        "created_at",
+        "issuer",
+    )
+    readonly_fields = (
+        "created_at",
+        "is_expired",
+        "issuer",
+    )
+    list_display = (
+        "email",
+        "team",
+        "created_at",
+        "is_expired",
+    )
+
+    def save_model(self, request, obj, form, change):
+        obj.issuer = request.user
+        obj.save()
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -148,3 +148,12 @@ class TeamSerializer(serializers.ModelSerializer):
     def get_slug(self, instance):
         """Return slug from the team's name."""
         return instance.get_slug()
+
+
+class InvitationSerializer(serializers.ModelSerializer):
+    """Serialize invitations."""
+
+    class Meta:
+        model = models.Invitation
+        fields = ["email", "team", "role", "issuer"]
+        read_only_fields = ["team", "issuer"]
--- a/src/backend/core/factories.py
+++ b/src/backend/core/factories.py
@@ -170,3 +170,15 @@ class TeamAccessFactory(factory.django.DjangoModelFactory):
     team = factory.SubFactory(TeamFactory)
     user = factory.SubFactory(UserFactory)
     role = factory.fuzzy.FuzzyChoice([r[0] for r in models.RoleChoices.choices])
+
+
+class InvitationFactory(factory.django.DjangoModelFactory):
+    """A factory to create invitations for a user"""
+
+    class Meta:
+        model = models.Invitation
+
+    email = factory.Faker("email")
+    team = factory.SubFactory(TeamFactory)
+    role = factory.fuzzy.FuzzyChoice([role[0] for role in models.RoleChoices.choices])
+    issuer = factory.SubFactory(UserFactory)
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -4,12 +4,14 @@ Declare and configure the models for the People core application
 import json
 import os
 import uuid
+from datetime import timedelta
 
 from django.conf import settings
 from django.contrib.auth import models as auth_models
 from django.contrib.auth.base_user import AbstractBaseUser
 from django.core import exceptions, mail, validators
 from django.db import models
+from django.utils import timezone
 from django.utils.functional import lazy
 from django.utils.text import slugify
 from django.utils.translation import gettext_lazy as _
@@ -451,6 +453,34 @@ class TeamAccess(BaseModel):
             "put": bool(set_role_to),
             "set_role_to": set_role_to,
         }
+
+
+class Invitation(BaseModel):
+    """User invitation to teams."""
+
+    email = models.EmailField(_("email address"), null=False, blank=False)
+    team = models.ForeignKey(
+        Team,
+        on_delete=models.CASCADE,
+        related_name="invitations",
+    )
+    role = models.CharField(
+        max_length=20, choices=RoleChoices.choices, default=RoleChoices.MEMBER
+    )
+    issuer = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+        related_name="invitations",
+    )
+
+    def __str__(self):
+        return f"{self.email} invited to {self.team}"
+
+    @property
+    def is_expired(self):
+        """Calculate if invitation is still valid or has expired."""
+        validity_duration = timedelta(seconds=settings.INVITATION_VALIDITY_DURATION)
+        return timezone.now() > (self.created_at + validity_duration)
 
 
 def oidc_user_getter(validated_token):
--- a/src/backend/core/tests/test_models_invitations.py
+++ b/src/backend/core/tests/test_models_invitations.py
@@ -1,0 +1,46 @@
+"""
+Unit tests for the Invitation model
+"""
+from django.core.exceptions import ValidationError
+
+import pytest
+
+from core import factories
+
+pytestmark = pytest.mark.django_db
+
+
+def test_models_invitations_email_no_empty_mail():
+    """The "email" field should not be empty."""
+    with pytest.raises(ValidationError, match="This field cannot be blank"):
+        factories.InvitationFactory(email="")
+
+
+def test_models_invitations_email_no_null_mail():
+    """The "email" field is required."""
+    with pytest.raises(ValidationError, match="This field cannot be null"):
+        factories.InvitationFactory(email=None)
+
+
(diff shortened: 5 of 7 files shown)


src/backend/core/models.py
```diff
@@ -452,6 +452,27 @@ def get_abilities(self, user):
         }
 
 
+class Invitation(BaseModel):
```
lebaudantoine, line 455: NIceeee, I would suggest adding
- A foreign key to the sender/issuer of the invitation.
- A validity duration, with a default value of 7 days? Or an expiration date

So we match Github invitation: 
```
@[mjeammet](https://github.com/mjeammet) has invited you to join the @numerique-gouv organization on GitHub. Head over to https://github.com/numerique-gouv to check out @numerique-gouv’s profile.

This invitation will expire in 7 days.
```

lebaudantoine:
Should we register this newly created model to the admin?

src/backend/core/models.py
mjeammet: Hey, cool proposals ! About the sender/issuer, its pretty straightforward. For the validity duration, should I actually transform `is_pending` to a `status` field, with three choices : "pending", "expired", "archived/resolved" ?

lebaudantoine: I am pretty sure we can challenge the `is_pending` boolean field. As now, `is_pending` can not be falsy. If a user accepts or rejects the invitation, we would delete the row, meaning that only rows with ìs_pending === true` would be stored in this table. IMO, this attribute as boolean doesn't provide much value, but I might be wrong, and forget some cases.

mjeammet: OK ! So, as discuted on tchap : 

- `is_pending` is removed. Invitation status will be calculated via a property 
- expired invitations are kept in database, to allow for re-invitations

mjeammet: In this model, accepted invitations are deleted and converted to team accesses.

src/backend/core/api/serializers.py
```diff
@@ -148,3 +148,10 @@ def get_abilities(self, team) -> dict:
     def get_slug(self, instance):
         """Return slug from the team's name."""
         return instance.get_slug()
+
+class InvitationSerializer(serializers.ModelSerializer):
+    team = TeamSerializer()
```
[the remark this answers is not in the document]
mjeammet: I don't quite remember how that works. I applied your suggestion and will figure it out in the next PR.

src/backend/core/api/serializers.py
```diff
@@ -148,3 +148,10 @@ def get_abilities(self, team) -> dict:
...
+
+    class Meta:
+        model = models.Invitation
+        fields = ["email", "team", "role", "is_pending"]
```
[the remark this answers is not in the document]
mjeammet: I removed it altogether

src/backend/core/models.py
```diff
@@ -452,6 +454,34 @@ def get_abilities(self, user):
...
+    team = models.ForeignKey(
+        Team,
+        on_delete=models.CASCADE,
+        related_name="invitations",
+    )
+    role = models.CharField(
+        max_length=20, choices=RoleChoices.choices, default=RoleChoices.MEMBER
+    )
+    issuer = models.ForeignKey(
+        User, on_delete=models.SET_NULL, related_name="invitations", null=True
```
mjeammet, line 470: I set `on_delete` to `SET_NULL` == If a user delete their account, their pending invitations should remain valid. 

Do you agree ?

src/backend/core/models.py
```diff
@@ -452,6 +454,34 @@ def get_abilities(self, user):
...
+    )
+
+    def __str__(self):
+        return f"{self.email} invited to {self.team}"
+
+    @property
+    def is_pending(self):
+        """Calculate if invitation is still valid or has expired."""
+        tz = pytz.timezone(settings.TIME_ZONE)
+        return datetime.utcnow(tz=tz) < (
```
mjeammet, line 480: I had to grab the project timezone in the expected format. If there is a cleaner way to do it, feel free to suggest it.

src/backend/people/settings.py
```diff
@@ -280,6 +281,7 @@ class Base(Configuration):
     EMAIL_FROM = values.Value("from@example.com")
 
     AUTH_USER_MODEL = "core.User"
+    INVITATION_VALIDITY_DURATION = timedelta(days=7)
```
mjeammet, line 284: I added it as timedelta to be able to import it directy whenever needed. Still very readable and easy to modify.

src/backend/core/models.py
sampaccoud: If a user is deleted, I would delete the invitations they made... because it's weird to invite someone to a group you're leaving shortly afterwards?

src/backend/core/api/serializers.py
```diff
@@ -148,3 +148,11 @@ def get_abilities(self, team) -> dict:
...
         """Return slug from the team's name."""
         return instance.get_slug()
+
+
+class InvitationSerializer(serializers.ModelSerializer):
+    """Serialize invitations."""
+
+    class Meta:
+        model = models.Invitation
+        fields = ["email", "team", "role", "issuer"]
```
sampaccoud, line 158: issuer should be readonly and set programmatically from the logged-in user?

src/backend/core/models.py
```diff
@@ -452,6 +454,34 @@ def get_abilities(self, user):
...
+    )
+    role = models.CharField(
+        max_length=20, choices=RoleChoices.choices, default=RoleChoices.MEMBER
+    )
+    issuer = models.ForeignKey(
+        User, on_delete=models.SET_NULL, related_name="invitations", null=True
+    )
+
+    def __str__(self):
+        return f"{self.email} invited to {self.team}"
```
sampaccoud, line 474: should we translate it?

src/backend/core/models.py
```diff
@@ -452,6 +454,34 @@ def get_abilities(self, user):
...
+
+    @property
+    def is_pending(self):
```
sampaccoud, line 477: I would say it is easier to understand a field called `is_expired`?

src/backend/core/models.py
sampaccoud: If it's still a good practice, you can use:
```django
from django.utils import timezone
now = timezone.now()
```

src/backend/core/tests/test_models_invitations.py
```diff
@@ -0,0 +1,46 @@
...
+
+import pytest
+
+from core import factories
+
+pytestmark = pytest.mark.django_db
+
+
+def test_models_invitations_email_no_empty_mail():
+    """The "email" field is required."""
```
sampaccoud, line 14: Maybe to differentiate from the next test's docstring:
```suggestion
    """The "email" field should not be empty."""
```

src/backend/people/settings.py
sampaccoud: I think it should be a duration in seconds so it can be set via environment variable. 
The timedelta can be instantiated on the use site with something like:
```
instattion_validity_duration = timedelta(seconds=settings.INVITATION_VALIDITY_DURATION)
```

src/backend/core/factories.py
```diff
@@ -170,3 +170,15 @@ class Meta:
...
+class InvitationFactory(factory.django.DjangoModelFactory):
+    """A factory to create invitations for a user"""
+
+    class Meta:
+        model = models.Invitation
+
+    email = factory.Faker("email")
+    team = factory.SubFactory(TeamFactory)
+    role = factory.fuzzy.FuzzyChoice([role[0] for role in models.RoleChoices.choices])
+    issuer = factory.SubFactory(UserFactory)
```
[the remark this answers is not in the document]
sampaccoud: not only belong but have administrator or owner role.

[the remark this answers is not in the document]
sampaccoud: maybe we can check that upon creating the invitation and upon consuming the invitation as well?

src/backend/core/api/serializers.py
mjeammet: I was unsure how and where to do that (serializer ? viewset ?). I finally opted for [this method](https://profil-software.com/blog/development/10-things-you-need-know-effectively-use-django-rest-framework/#using-currentuserdefault).

src/backend/core/models.py
mjeammet: What do you mean ? Won't we translate it in crowdin ?

src/backend/core/api/serializers.py
mjeammet: I think we could handle this in the PR #50, where I create the route and the related tests.

lebaudantoine: here ? https://github.com/numerique-gouv/people/pull/50/files#diff-78bd491985b92670ff5ca54200235d6db2810a6a1a999e673392a90a86e86c6fR159

src/backend/core/factories.py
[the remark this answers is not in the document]
mjeammet: Done ! I added that to #50

src/backend/people/settings.py
mjeammet: Done !

src/backend/core/models.py
mjeammet: Done !

src/backend/core/admin.py
```diff
@@ -110,3 +110,29 @@ class TeamAdmin(admin.ModelAdmin):
...
+
+@admin.register(models.Invitation)
+class InvitationAdmin(admin.ModelAdmin):
+    """Admin interface to handle invitations."""
+
+    fields = (
+        "email",
+        "team",
+        "role",
+        "created_at",
```
lebaudantoine, line 123: nit: missing issuer ?

lebaudantoine (approved):
A minor comment, LGTM ship it.

src/backend/core/admin.py
mjeammet: Was added in the next PR, had forgotten to bring it back here. It's dooone !

PR #50, opened 2024-02-12
✨(backend) add invitation CRUD

mjeammet:
## Purpose

Add an invitation router, nested below teams/ 
- `create` to allow team owners/administrators to invite members
- `list` pending and expired invitations
- `delete` to cancel invitations

## Proposal

- [x] add route and viewset
- [x] add create action
- [x]  add uniqueness together to avoid duplicate of email in the same team
- [x] permission tests
- [x] list 
- [x] tests list
- [x] test expired invitations are listed too
- [x] retrieve
- [x] tests retrieve
- [x] block update in admin
- [x] delete endpoint 
- [x] tests delete
- [x] added validations on invitation.email and related tests

Retrieve and update actions will be dealt with later.

## Unrelated fixes 

- Fixed access permissions of anonymous user and team-unrelated user
- Fixed warning occurring a lot during tests


src/backend/core/models.py
```diff
@@ -452,6 +454,41 @@ def get_abilities(self, user):
...
+        on_delete=models.CASCADE,
+        related_name="invitations",
+    )
+
+    class Meta:
+        constraints = [
+            models.UniqueConstraint(
+                fields=["email", "team"], name="email_and_team_unique_together"
+            )
+        ]
```
mjeammet, line 480: I added uniqueness together to prevent duplication of invitation to the same team. 

An email can be invited to several teams though, so we'll have to decide if we send an invitation mail each time or not.

src/backend/core/tests/teams/test_core_api_teams_list.py
mjeammet, line 1: Looks like this file was previously ignored by the linter.

src/backend/core/tests/test_api_team_invitations.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+@pytest.mark.parametrize(
+    "parameters",
+    [
+        ("owner", status.HTTP_201_CREATED),
+        ("administrator", status.HTTP_201_CREATED),
+        ("member", status.HTTP_403_FORBIDDEN),
+    ],
+)
```
mjeammet, line 60: I really love this feature 😍

src/backend/core/tests/teams/test_core_api_teams_list.py
mjeammet: It reminds me that we have to decide if people outside of team can see it / can list its members. Should I open an issue ?

src/backend/core/api/viewsets.py
```diff
@@ -376,3 +377,38 @@ def perform_update(self, serializer):
...
+    mixins.CreateModelMixin,
+    viewsets.GenericViewSet,
+):
+    """API ViewSet for user invitations to team.
+
+    POST /api/v1.0/teams/<team_id>/invitations/ with expected data:
+        - email: str
+        - role: str [owner|admin|member]
+        - issuer : User, automatically added from user making query, if allowed
+        - team : Team, automatically added from requested URI
```
sampaccoud, line 392: Maybe shouldn't be listed here since this is "expected data" so the expected request payload?

src/backend/core/models.py
sampaccoud: There should be a diff for the migration file as well?

src/backend/core/api/viewsets.py
```diff
@@ -376,3 +377,38 @@ def perform_update(self, serializer):
...
+    queryset = models.Invitation.objects.all().select_related("team")
+    serializer_class = serializers.InvitationSerializer
+
+    def perform_create(self, serializer, *args, **kwargs):
+        """Verify the issuer is allowed to send invitations to this team."""
+        user = self.request.user
+        team = get_object_or_404(models.Team, pk=self.kwargs["team_id"])
+
+        # check user is team's owner or administrator
+        if not team.get_abilities(user)["manage_accesses"]:
```
sampaccoud, line 408: Maybe better placed in a permission which you can add to the "permissions" attribute above like:
```python
permission_classes = [permissions.AccessPermission, permissions.RelatedTeamAccessPermission]
```

sampaccoud (commented):
Nice one :star_struck:

src/backend/core/models.py
mjeammet: You're absolutely right, I had forgotten to do it again

src/backend/core/api/viewsets.py
mjeammet, line 392:
```suggestion
    POST /api/v1.0/teams/<team_id>/invitations/ with expected data:
        - email: str
        - role: str [owner|admin|member]
        issuer and team are added automatically from request user and URI
```
Hm, you're right. Is this better ?

mjeammet:
```suggestion
        (issuer and team are automatically added, from request user and URI)
```
You're right. Is this better ? I feel like this modification should be explicited in docs

src/backend/core/api/viewsets.py
mjeammet: Done ! Took me a hot minute, I couldn't remember how permissions classes worked.

src/backend/core/api/permissions.py
```diff
@@ -52,3 +54,23 @@ def has_object_permission(self, request, view, obj):
         """Check permission for a given object."""
         abilities = obj.get_abilities(request.user)
         return abilities.get(request.method.lower(), False)
+
+
+class RelatedTeamAccessPermission(IsAuthenticated):
```
sampaccoud, line 59: I looked into it and now I think we shouldn't merge this PR without PR https://github.com/numerique-gouv/people/pull/65.
In fact this `invitation` endpoint should work exactly like the `team-access` endpoint. The filtering of accesses related to a team can be done the `get_queryset` method without incurring extra queries. See these lines in the TeamAccessViewset:
```

    def get_queryset(self):
        """Return the queryset according to the action."""
        queryset = super().get_queryset()
        queryset = queryset.filter(team=self.kwargs["team_id"])
```

mjeammet: OK. What I was suggesting the other day was to 
- merge this PR here quickly to allow Anto to work with this creating endpoint
- keep on advancing on #65 to complete the CRUD soon and handle listing / deletion soon

I can modify this PR to make it match team accesses but I still think we should merge it quickly and improve the finer details regarding the whole endpoint on next PR.

mjeammet:
Hi @AntoLC ! I'm not sure what remains to be added for that PR to fit your needs.

There's a list in the PR description, do you see anything that you might want me to add to it ?

src/backend/core/tests/test_api_team_invitations.py
```diff
@@ -0,0 +1,299 @@
...
+
+    valid_invit = factories.InvitationFactory(
+        team=team, role="administrator", issuer=identity.user
+    )
+
+    client = APIClient()
+    client.force_login(identity.user)
+    response = client.get(
+        f"/api/v1.0/teams/{team.id}/invitations/",
+    )
```
mjeammet, line 224: - I can't set up `created_at` easily, 
- changing INVITATION_VALIDITY_DURATION impact both invitations
- I didn't want to use an additional package (such as [freezegun](https://github.com/spulec/freezegun/blob/master/README.rst) for instance) to avoid an extra dependancy solely for this test 

so I chose a super naive approach. I suppose, in theory, on very old machines, test could fail because `valid_invitation` becomes invalid in the time we query the endpoint but I don't think I could really happen neither on local nor on our servers.

src/backend/core/models.py
```diff
@@ -462,11 +462,20 @@ class Invitation(BaseModel):
...
+
     def __str__(self):
         return f"{self.email} invited to {self.team}"
 
     @property
     def is_expired(self):
         """Calculate if invitation is still valid or has expired."""
         validity_duration = timedelta(seconds=settings.INVITATION_VALIDITY_DURATION)
+        if not self.created_at:
+            return False
```
mjeammet, line 480: returns False when object is not yet instanciated ? I'm unsure.

mjeammet:
@sampaccoud @AntoLC @lebaudantoine

src/backend/core/tests/test_api_team_invitations.py
mjeammet, line 1: I changed naming for this file. WDYT ?

src/backend/core/factories.py
```diff
@@ -148,6 +148,7 @@ class TeamFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = models.Team
         django_get_or_create = ("name",)
+        skip_postgeneration_save = True
```
lebaudantoine, line 151: could you give more context on this one?

src/backend/core/models.py
lebaudantoine, line 480: I would suggest an early return :
```suggestion
    def is_expired(self):
        """Calculate if invitation is still valid or has expired."""
        if not self.created_at:
            return False
        validity_duration = timedelta(seconds=settings.INVITATION_VALIDITY_DURATION)
```

lebaudantoine: Why did you choose `False` over `None`?

src/backend/core/api/serializers.py
```diff
@@ -228,5 +228,5 @@ class InvitationSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = models.Invitation
-        fields = ["email", "team", "role", "issuer"]
-        read_only_fields = ["team", "issuer"]
+        fields = ["created_at", "email", "team", "role", "issuer", "is_expired"]
+        read_only_fields = ["created_at", "team", "issuer", "is_expired"]
```
lebaudantoine, line 232:
```suggestion
        fields = ["id", "email", "team", "role", "issuer", "is_expired", "created_at"]
        read_only_fields = ["id", "team", "issuer", "is_expired", "created_at"]
```

lebaudantoine: we need the ID to perform delete operation.

src/backend/core/api/viewsets.py
```diff
@@ -399,3 +400,80 @@ def perform_update(self, serializer):
                 raise exceptions.ValidationError({"role": message})
 
         serializer.save()
+
+
+class InvitationViewset(
```
lebaudantoine, line 405: Should I be allowed to update an invitation in the admin ?

Capture d’écran 2024-03-05 à 12 36 13

src/backend/core/api/viewsets.py
```diff
@@ -399,3 +400,80 @@ def perform_update(self, serializer):
...
+    mixins.CreateModelMixin,
```
lebaudantoine, line 406: We might need additional validation checks. ⚠️

When generating invitations, I shouldn't be able to:
- Invite someone using an email already associated with a team member.
- Invite someone using an email linked to an existing user.

Let's enhance our validation to ensure a smoother process!

src/backend/core/api/permissions.py
```diff
@@ -50,5 +53,21 @@ class AccessPermission(IsAuthenticated):
 
     def has_object_permission(self, request, view, obj):
         """Check permission for a given object."""
+        if isinstance(obj, models.Invitation):
+            team = get_object_or_404(models.Team, id=view.kwargs["team_id"])
+            return team.get_abilities(request.user).get("manage_accesses")
+
         abilities = obj.get_abilities(request.user)
         return abilities.get(request.method.lower(), False)
```
lebaudantoine, line 61: It feels weird introducing a custom logic using an if in a generic classes used by all the models. Could we extend the  default classes, to respect the [open-close principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle)

lebaudantoine: it seems duplicated with the `RelatedTeamAccessPermission` classes that inherit from `IsAuthenticated`

src/backend/core/api/permissions.py
```diff
@@ -50,5 +53,21 @@ class AccessPermission(IsAuthenticated):
...
+
+
+class RelatedTeamAccessPermission(IsAuthenticated):
```
lebaudantoine, line 64: Naming could be improved, to `RelatedTeamPermission` or something else. Reading it first, it feels the class has the responsibility to handle permissions for `TeamAccesses` only, when the doc states it's for team accesses and invitations.

src/backend/core/api/permissions.py
```diff
@@ -50,5 +53,21 @@ class AccessPermission(IsAuthenticated):
...
+    """Permission class for team accesses and invitations."""
+
+    message = "Only administrators can invite members."
```
lebaudantoine, line 67: should it be `"only owner or administrators can invite members"`?

src/backend/people/api_urls.py
```diff
@@ -21,6 +21,11 @@
     basename="team_accesses",
 )
```
lebaudantoine, line 23: could we get inspiration from [here](https://github.com/openfun/joanie/blob/538d737f2f4090a7aac36df694534178b2cd3181/src/backend/joanie/client_urls.py#L57C1-L57C22).
to avoid duplicating the `re_path` and have a unique and dedicated router for team related routes.

src/backend/core/tests/test_api_team_invitations.py
lebaudantoine: I would recommend you to add [freezegun](https://github.com/spulec/freezegun/blob/master/README.rst)

lebaudantoine: Dates are a complex thing, and expiration data a critical part. We should test them properly, feel free to use the appropriate tool for it.

lebaudantoine: or, you could try to mock the default function that give created_at its value. I am not sure it would work.

lebaudantoine (commented):
I haven't read the tests. Thx for this huge contributions, some major and minor comments here 🙏 we're close to shipppp it hereeee ❤️

src/backend/core/api/serializers.py
AntoLC: Yes, needed

src/backend/core/factories.py
mjeammet: I wanted to fix a warning occurring 124 times when running back end test suite : 
```
/usr/local/lib/python3.10/site-packages/factory/django.py:181: DeprecationWarning: TeamFactory._after_postgeneration will stop saving the instance after postgeneration hooks in the next major release.
  If the save call is extraneous, set skip_postgeneration_save=True in the TeamFactory.Meta.
  To keep saving the instance, move the save call to your postgeneration hooks or override _after_postgeneration.
    warnings.warn(
```

AntoLC (commented):
Nice, all the CRD works well, just need the invitationId in the list. 
I don't know if we need the U update, we cannot update the role by example if an invitation is already created, maybe it is fine like that.

I have a small question, why the tablename is `core_invitation` instead of `people_invitation` ?

src/backend/core/api/serializers.py
mjeammet: Done !

src/backend/people/api_urls.py
mjeammet: I didn't think that through, it's a good idea !

src/backend/core/models.py
mjeammet: Well, I'm not even sure if it could happen in a real situation, I wrote that when testing `is expired` before saving object. I'm still leaving it, in case it occurs in another situation.

No objection whatsoever to return "None" instead of "False".

mjeammet:
@AntoLC to answer your review comment : 
- thanks 😁 !

- update : I think we agreed to not update invitations for now. We can still decide to add it in V2 if we decide it's useful ! 

- tablename `core_invitation` : tt's the default because I didn't specify anything in the model. Should I ?

mjeammet:
@AntoLC thanks for your review ! To answer your review comment 


> Nice, all the CRD works well, just need the invitationId in the list. I don't know if we need the U update, we cannot update the role by example if an invitation is already created, maybe it is fine like that.

We'd agreed to not update invitations for now. We can still decide to add it in V2 if we decide it's useful ! 

> I have a small question, why the tablename is `core_invitation` instead of `people_invitation` ?

I's the default because I didn't specify anything in the model. Should I ?

src/backend/core/api/viewsets.py
mjeammet: I don't think so. I'll make all these fields `read_only` in the admin.

mjeammet: ✅ ! I marked all fields as read_only if the object already exists, and removed save buttons. I made sure we can still delete objects

I stole everything from this thread https://stackoverflow.com/questions/4343535/django-admin-make-a-field-read-only-when-modifying-obj-but-required-when-adding

src/backend/core/api/permissions.py
sampaccoud: Would it be too smart if we test:
```python
if view.kwargs.get("team_id")
```
If we have this kwarg, it means that we are in an url nested under a team? Seems ok and generic to me with a comment to explain.

src/backend/core/tests/test_api_team_invitations.py
sampaccoud: you added "team" because it's nested is it the change? Seems neet :rocket:

mjeammet: More like "section__action__context". But since "section" is already the name of the file, "action__context" would do.

src/backend/core/api/permissions.py
mjeammet: I'd like to keep a dedicated permission for Invitation, mostly for the custom error message (and the linter doesn't appreciate the error message being anywhere else than at class level). 

I'll modify what I need in InvitationPermission then, as it seem that TeamAccesses don't have the same problem 🤔

src/backend/core/api/viewsets.py
```diff
@@ -399,3 +400,80 @@ def perform_update(self, serializer):
...
+    """
+
+    lookup_field = "id"
+    pagination_class = Pagination
+    permission_classes = [
+        permissions.AccessPermission,
+        permissions.RelatedTeamAccessPermission,
+    ]
+    queryset = (
+        models.Invitation.objects.all().select_related("team").order_by("-created_at")
```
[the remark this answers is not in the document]
mjeammet: I don't know. @sampaccoud ?

src/backend/core/tests/test_api_team_invitations.py
mjeammet: I realized what we really wanted to test here is that expired invitations don't disappear and are still listed correctly. I don't need to tweak spacetime to do that. I slightly modified the test and it's robust now, I don't think I need to add freezegun or anything else.

src/backend/core/api/viewsets.py
```diff
@@ -313,7 +314,7 @@ class TeamAccessViewSet(
     lookup_field = "pk"
     pagination_class = Pagination
     permission_classes = [permissions.AccessPermission]
-    queryset = models.TeamAccess.objects.all().select_related("user")
+    queryset = models.TeamAccess.objects.all().select_related("user").order_by("role")
```
lebaudantoine, line 317: What type of warning were raised?

src/backend/core/api/viewsets.py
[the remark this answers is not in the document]
lebaudantoine: also, i'll open an issue, but it seems that pagination uses `created_on` and, it doesn't seem to match any attribute of the base model (fyi @mjeammet)

src/backend/core/api/serializers.py
```diff
@@ -228,5 +230,24 @@ class InvitationSerializer(serializers.ModelSerializer):
...
+            if existing_access:
+                raise serializers.ValidationError(
+                    "This member already joined your team."
+                )
+
+            raise serializers.ValidationError(
+                _("This email is already associated to a registered user.")
+            )
+
+        return attrs
```
lebaudantoine, line 253: nit early return to avoid nesting. I am a bit scared to have introduced regression, is this logic identical with your original code?

Also, I have added `.exists()` to the end of filters; Which should be a bit more performant:

> simply reads a single record in [the most optimized way](https://github.com/django/django/blob/b996424199cca58f64b77faeaa367b7e7efb9776/django/db/models/sql/query.py#L528):


```suggestion
    def validate(self, attrs):
    
        existing_identity = models.Identity.objects.filter(email=attrs["email"]).exists()
        
        # Verify invitee has not already joined the app / the team
        if not existing_identity:
                return attrs
                
        existing_access = models.TeamAccess.objects.filter(
                team=self.initial_data["team"]
        ).exists()
        
        if existing_access:
                raise serializers.ValidationError(
                        "This member already joined your team."
                )

        raise serializers.ValidationError(
                _("This email is already associated to a registered user.")
        )

```

lebaudantoine: we could add a docstring, strange the linter doesn't fail

lebaudantoine (commented):
THHHHHx 
we almost there

src/backend/core/api/serializers.py
```diff
@@ -228,5 +230,46 @@ class InvitationSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = models.Invitation
-        fields = ["email", "team", "role", "issuer"]
-        read_only_fields = ["team", "issuer"]
+        fields = ["id", "created_at", "email", "team", "role", "issuer", "is_expired"]
+        read_only_fields = ["id", "created_at", "team", "issuer", "is_expired"]
+
+    def validate(self, attrs):
```
lebaudantoine, line 236: @sampaccoud 👀

src/backend/core/models.py
```diff
@@ -462,11 +462,49 @@ class Invitation(BaseModel):
...
     @property
     def is_expired(self):
         """Calculate if invitation is still valid or has expired."""
+        if not self.created_at:
+            return None
+
         validity_duration = timedelta(seconds=settings.INVITATION_VALIDITY_DURATION)
         return timezone.now() > (self.created_at + validity_duration)
+
+    def get_abilities(self, user):
```
lebaudantoine, line 487: @sampaccoud 👀

src/backend/core/api/viewsets.py
```diff
@@ -186,7 +186,7 @@ class UserViewSet(
     """
 
     permission_classes = [permissions.IsSelf]
-    queryset = models.User.objects.all()
+    queryset = models.User.objects.all().order_by("-created_at")
```
lebaudantoine, line 189: I wonder, if by default we should not define an ordering field in the `BaseModel`, might solve all warnings

sampaccoud: good idea since we have this field on all models!

sampaccoud (approved):
I proposed some slight simpilfications:
- validate directly on the model constraints that should also apply to the model
- don't need to check if a TeamAccess already exists... because if it did, the Identity would also exist anyway and the first error would already be raised...

lebaudantoine:
> don't need to check if a TeamAccess already exists... because if it did, the Identity would also exist anyway and the first error would already be raised...

Yes, but we might handle it in different way in frontend :
- Existing identity only → re-do the request, convert the invitation to team access
- Already member of the team → raise an error

src/backend/core/admin.py
```diff
@@ -144,6 +144,20 @@ class InvitationAdmin(admin.ModelAdmin):
         "is_expired",
     )
 
+    def get_readonly_fields(self, request, obj=None):
+        if obj:
```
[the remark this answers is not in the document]
lebaudantoine: https://github.com/numerique-gouv/people/issues/104

src/backend/core/api/viewsets.py
lebaudantoine: will do next

PR #53, opened 2024-02-13
🎨(frontend) Consistency url

AntoLC:
## Purpose

We have this kind of url `/teams/1321465`, but `/teams/` gives a 404, it is not a good practice.
We make `/teams` as an alias of the homepage `/`.

## Proposal

- [x] create teams page as alias of home
- [x] test

## Demo 

[scrnli_2_13_2024_12-03-10 PM.webm](https://github.com/numerique-gouv/people/assets/25994652/6d637779-9574-47ee-83d8-90528303677f)

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-desk/teams-create.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-desk/teams-create.spec.ts
@@ -80,4 +80,18 @@ test.describe('Teams Create', () => {
     await panel.locator('li').getByText(teamName).click();
     await expect(elTeam).toBeVisible();
   });
+
+  test('checks alias teams url with homepage', async ({ page }) => {
+    await expect(page).toHaveURL('/');
+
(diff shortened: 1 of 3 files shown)


src/frontend/apps/e2e/__tests__/app-desk/teams-create.spec.ts
```diff
@@ -80,4 +80,16 @@ test.describe('Teams', () => {
     await panel.locator('li').getByText(teamName).click();
     await expect(elTeam).toBeVisible();
   });
+
+  test('checks alias teams url with homepage', async ({ page }) => {
+    const buttonCreateHomepage = page.getByRole('button', {
```
lebaudantoine, line 85: nit:
```suggestion
    await expect(page).toHaveURL(/);
```
What do you think ?
It makes explicit the test assert the `/` and `/teams` share the same content

lebaudantoine (approved):

PR #58, opened 2024-02-14
🩹(backend) fix Django Admin UserAdmin page

### LINKED ISSUE #56: Can't create users via admin
## Bug Report

**Problematic behavior**
Cannot create new user in admin due to lack of username.

**Expected behavior/code**
I'd like Django to expect an email and not a username.

**Steps to Reproduce**
1. `make superuser` to login to admin
2. navigate to users admin
3. try to add a new user

**Environment**
- People version: dev 7ff00a9401af3db3062a7763dba44f8d9dbf2291 
- Platform: linux

**Possible Solution**

**Additional context/Screenshots**
image

```
FieldError at /admin/core/user/add/
Unknown field(s) (username) specified for User. Check fields/fieldsets/exclude attributes of class UserAdmin.
```

lebaudantoine:
Fix several issues identified in the Django Admin

 **Broken Identity string representation**
Resolving a format error in the Identity string representation caused by potential `None` values in the email field. This issue was discovered when attempting to access the User details page in the Django Admin

**Broken User creation form**
The replacement of the User's username with an email led to errors in the `UserAdmin` class. The base class used the `username` field in the `add_fieldsets` attribute. This problem was discovered while attempting to create a new user in the Django Admin.

It closes #56.

### DIFF
--- a/src/backend/core/admin.py
+++ b/src/backend/core/admin.py
@@ -82,6 +82,15 @@ class UserAdmin(auth_admin.UserAdmin):
         ),
         (_("Important dates"), {"fields": ("created_at", "updated_at")}),
     )
+    add_fieldsets = (
+        (
+            None,
+            {
+                "classes": ("wide",),
+                "fields": ("email", "password1", "password2"),
+            },
+        ),
+    )
     inlines = (IdentityInline, TeamAccessInline)
     list_display = (
         "email",
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -288,7 +288,8 @@ class Identity(BaseModel):
 
     def __str__(self):
         main_str = "[main]" if self.is_main else ""
-        return f"{self.email:s}{main_str:s}"
+        id_str = self.email or self.sub
+        return f"{id_str:s}{main_str:s}"
 
     def save(self, *args, **kwargs):
         """Ensure users always have one and only one main identity."""


src/backend/core/admin.py
```diff
@@ -82,6 +82,15 @@ class UserAdmin(auth_admin.UserAdmin):
         ),
         (_("Important dates"), {"fields": ("created_at", "updated_at")}),
     )
+    add_fieldsets = (
+        (
+            None,
+            {
+                "classes": ("wide",),
```
mjeammet, line 89: Isn't that a style attribute ? I'm curious : why add it here ?

src/backend/core/models.py
```diff
@@ -288,7 +288,8 @@ class Meta:
 
     def __str__(self):
         main_str = "[main]" if self.is_main else ""
-        return f"{self.email:s}{main_str:s}"
+        id_str = self.email or self.sub
```
mjeammet, line 291: I would have expected it to fix the display in the admin, but I still have "-" when no email are given 🤔

mjeammet (approved):
👍🏽

src/backend/core/admin.py
lebaudantoine: I overrode default values from the base class [here](https://github.com/django/django/blob/3cadeea077a98367a4ed344d645df0aff243de91/django/contrib/auth/admin.py#L64). I replaced the field username by email, and kept others values as it was. Wdyt?

mjeammet: Smart ! OK, thanks then, curiosity satiated.

PR #60, opened 2024-02-15
🗃️(backend) first name and last name to identity

AntoLC:
## Purpose

We need a name to display team members in the frontend.

## Proposal

- [x] add colum `name` on `people_identity` table
- [x] sync with Keycloak
- [x] tests


src/backend/people/settings.py
```diff
@@ -334,6 +333,12 @@ class Base(Configuration):
         default=[], environ_name="OIDC_REDIRECT_ALLOWED_HOSTS", environ_prefix=None
     )
 
+    USER_OIDC_FIELDS_TO_NAME = values.ListValue(
+        default=["first_name", "last_name"],
```
lebaudantoine, line 337: I have quickly checked the documentation, fyi agent connect exposes `given_name` and `usual_name` in their token I think based [on the documentation ](https://github.com/france-connect/Documentation-AgentConnect/blob/main/doc_fs/projet_fca/projet_fca_donnees.md#les-donn%C3%A9es-obligatoires)

lebaudantoine: Never minds, as these are env values, they can be easily overrode. Nice move

lebaudantoine: nit: `USER_OIDC_FIELDS_TO_NAME` > `OIDC_USER_FIELDS_TO_NAME`

docker/auth/realm.json
```diff
@@ -46,6 +46,9 @@
   "users": [
     {
       "username": "people",
+      "email": "people@people.world",
+      "firstName": "John",
+      "lastName": "Doe",
```
lebaudantoine, line 51: Should we add similar mocked values on the other users used by e2e tests?

lebaudantoine (approved):

src/backend/people/settings.py
sampaccoud: ok! I chose the former because I wanted to leave OIDC_* as a namespace for the Mozilla third party application.  But maybe this is not necessary and what you propose is clearer to group it under OIDC matters.

docker/auth/realm.json
sampaccoud: ok. Can you point to where are those users?

lebaudantoine: Okay, then let's ignore my comment !

PR #62, opened 2024-02-15
✏️(project) fix minor typos

lebaudantoine:
Found typos, fixed them.


src/backend/core/tests/test_models_identities.py
```diff
@@ -142,7 +142,7 @@ def test_models_identities_sub_unique():
 
 
 def test_models_identities_sub_max_length():
-    """The sub field should be 255 characters maximum."""
+    """The subfield should be 255 characters maximum."""
```
mjeammet, line 145: That's unexpected. Why a single word ?

mjeammet (approved):

src/backend/core/tests/test_models_identities.py
lebaudantoine: mybad I'll fix it to `'sub' field`

src/backend/core/tests/test_models_identities.py
```diff
@@ -99,15 +99,15 @@ def test_models_identities_email_unique_different_users():
 
 
 def test_models_identities_email_normalization():
-    """The email field should be automatically normalized upon saving."""
+    """The "email" field should be automatically normalized upon saving."""
```
mjeammet, line 102: did you espace those `"` around "email" or are they different characters ?

src/backend/core/tests/test_models_identities.py
```diff
@@ -93,21 +93,21 @@ def test_models_identities_email_unique_same_user():
 
 
 def test_models_identities_email_unique_different_users():
-    """The "email" field should not be unique among users."""
+    """The 'email' field should not be unique among users."""
     email = factories.IdentityFactory()
     factories.IdentityFactory(email=email.email)
```
mjeammet, line 98: I know this is not the PR to raise this kind of questions but ... do we really want emails to not be unique among users ? 

The idea I imagine is to include mailing list ?

sampaccoud: Technically we don't need them to be unique as we rely on the sub... The email is just for sending emails. I didn't find any good reason why we should enforce this in the code. It could be troublesome in some scenarios and don't bring any good?

sampaccoud (approved):

PR #70, opened 2024-02-19
✨(backend) remove users returning profile_contact

mjeammet:
## Purpose

Custom UserManaged returned profile_contact field when returning users. While this may be useful later, we'd currently rather have it return users.

## Proposal

- [x] remove occurrences of User model returning `profile_contact`


src/backend/core/models.py
```diff
@@ -154,9 +154,9 @@ class UserManager(auth_models.UserManager):
     Override user manager to get the related contact in the same query by default (Contact model)
     """
 
-    def get_queryset(self):
+    def get_queryset(self):  # pylint: disable=useless-parent-delegation
         """Always select the related contact when doing a query on users."""
-        return super().get_queryset().select_related("profile_contact")
+        return super().get_queryset()
```
mjeammet, line 159: This custom UserManager has no reason to remain. Can I discard it, @sampaccoud ? We can re-add it if needed if desk and contact "merge"  again

sampaccoud: ok! why don't you remove the whole function override as it is now useless?

sampaccoud (approved):

PR #71, opened 2024-02-20
✨(api) improve team accesses endpoint

### LINKED ISSUE #64: Add user info nested in the team endpoint
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Add user info nested in the team endpoint

**Describe the solution you'd like**
- email
- firstname
- lastname

**
**Do you want to work on it through a Pull Request?**
yes

mjeammet:
## Purpose

Improve `/teams/<id>/accesses/` endpoint to return user info
Relies on #60. 
Fixes #64 

Currently, serializer returns : 
```
{
'id' : access.id
'user': access.user.id
'role': access.role
'abilitites': access.abilities
}
```
We want to 
- add user main identity's email
- add user main identity's first and last name
- remove user id ?

## Proposal

Description...

- [x] small commit to separate team accesses tests into diferent files.
- [x] return serialized userinfo in `/teams/<id>/accesses/`
- [x] tests
- [x] fix tests suite
- [x] optimize the number of queries ⚡ 

# Improvement

- [ ] improve permissions : have queries from unrelated users return 403


src/backend/core/api/serializers.py
```diff
@@ -26,6 +26,32 @@ def update(self, instance, validated_data):
...
+    class Meta:
+        model = models.Identity
+        fields = ["email", "first_name", "last_name"]
+        read_only_fields = ["email", "first_name", "last_name"]
+
+
+class UserListSerializer(serializers.ModelSerializer):
+    """Serialize users for list views, including nested serializers."""
+
+    main_identity = IdentitySerializer(read_only=True)
```
sampaccoud, line 41: I don't think you should expose the identity object to the API. 
The first_name of the main identity is the user's first name and is readonly so I would say we want to return:
```
{
    "id": "xyz",
    "user": {
        "id": "xyz",
        "email": "david.bowman@example.com",
        "first_name": "david",
        "last_name": "bowman",
    },
    "role": "member",
    "abilities": {...},
}
```
The "email" field will always come from the identity as users with an email set on the User model are filtered out of the API.

mjeammet: I agree, that's what I intended to do. I'm just making my way step by step.

src/backend/core/api/serializers.py
```diff
@@ -120,11 +167,33 @@ def validate(self, attrs):
...
+            "abilities",
+        ]
+
+
 class TeamSerializer(serializers.ModelSerializer):
     """Serialize teams."""
 
     abilities = serializers.SerializerMethodField(read_only=True)
-    accesses = TeamAccessSerializer(many=True, read_only=True)
+    # accesses = TeamAccessSerializer(many=True, read_only=True)
```
mjeammet, line 196: Remove accesses field returned upon calling `/teams/<id>/` view. Accesses are now available from `/teams/<id>/accesses/`.

src/backend/core/api/viewsets.py
```diff
@@ -322,22 +326,35 @@ def get_serializer_context(self):
...
+            return self.list_serializer_class
+        return self.detail_serializer_class
+
     def get_queryset(self):
         """Return the queryset according to the action."""
         queryset = super().get_queryset()
         queryset = queryset.filter(team=self.kwargs["team_id"])
 
-        if self.action == "list":
+        if self.action in ["list", "retrieve"]:
```
lebaudantoine, line 339: very-nit: Let's use the right data structure whenever it's possible. Here we could use a set. It's a good habit, that slightly improve performances in some situations.

```suggestion
        if self.action in {"list", "retrieve"}:
```

src/backend/core/api/serializers.py
```diff
@@ -26,22 +26,69 @@ def update(self, instance, validated_data):
...
+        """Return main identity's email."""
+        try:
+            return obj.main_identity[0].email
+        except TypeError:
+            return obj.main_identity.email
+        except IndexError:
+            main_identity = obj.identities.filter(is_main=True).first()
+            return obj.main_identity.email if main_identity else None
+        except AttributeError:
+            return None
```
lebaudantoine, line 71: Nit: these logics could be (a bit) refactored.

lebaudantoine (commented):
Hey, Nice job !!  😊
Amazing optimization 🚀

I had a few important remarks reading the code:

- For the sake of simplicity, could we keep a single `UserSerializer`? Is there a strong requirement for introducing `UserCompleteSerializer`? It feels it was made to optimize data sent to the client. Did I miss something?

- I wonder, should we be able to update TeamAccess's user? I remembered we discussed it, and discarded it for the sake of simplicity. If so, we could stick to a single TeamAccess serializer. Saving us a bit of logic that we won't maintain.

let's discuss it tomorrow

src/backend/core/api/serializers.py
mjeammet: Yes, I kinda wanted to make sure there was no secret reason to keep this field. But if you don't see any, let's remove it and we'll add it again later, if needed.

AntoLC (commented):
The endpoint bellow let the `name` and `email` at `null`, I suppose some tests can be improved on this endpoint:
http://localhost:8071/api/v1.0/users/?q=people
```json
{
    "count": 2,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": "8474550b-b27e-4283-b9dd-f0a905dbff23",
            "name": null,
            "email": null,
            "language": "en-us",
            "timezone": "UTC",
            "is_device": false,
            "is_staff": false
        },
        {
            "id": "6db6555f-378b-4bdc-8483-08dda9e65fe0",
            "name": null,
            "email": null,
            "language": "en-us",
            "timezone": "UTC",
            "is_device": false,
            "is_staff": false
        }
    ]
}
```

✅ The other endpoints work as expected.

src/backend/core/api/serializers.py
```diff
@@ -26,22 +26,69 @@ def update(self, instance, validated_data):
...
     class Meta:
         model = models.User
         fields = [
             "id",
+            "name",
             "email",
             "language",
             "timezone",
             "is_device",
             "is_staff",
```
mjeammet, line 45: @lebaudantoine We only need the first 3 fields for the teamaccess view. I chose to return only the necessary info and have a dedicated, shorter serializer (UserSerializer).

EDIT : I've found a way to have a single UserSerializer, but filter fields outside when calling the serializer. I'm pushing the code right away.

src/backend/core/api/serializers.py
mjeammet: It's a bit clumsy. Feel free to refacto if you see a cooler way to do it ! I don't

src/backend/core/api/serializers.py
```diff
@@ -26,22 +26,70 @@ def update(self, instance, validated_data):
         return super().update(instance, validated_data)
 
 
-class UserSerializer(serializers.ModelSerializer):
+class DynamicFieldsModelSerializer(serializers.ModelSerializer):
```
mjeammet, line 29: From https://www.django-rest-framework.org/api-guide/serializers/#dynamically-modifying-fields

lebaudantoine:
@AntoLC name and email can be null, ex: when creating a superuser, they don't have any identity at all, then these fields are null. We have to handle it in the frontend, because user login through SSO are only forced to have a valid sub. Email and Name can be null for whatever reason

lebaudantoine (approved):
Ship it

PR #86, opened 2024-03-04
🚧(api) order team accesses by role, email or name

mjeammet:
## Purpose

Allows users to order results of GET on team/<id>/accesses/ endpoint. This is a work in progress.

## Proposal

Description...

- [x] sort by role
- [] sort by email or name
- [x] dedicated test

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -2,6 +2,7 @@
 from django.contrib.postgres.search import TrigramSimilarity
 from django.db.models import Func, Max, OuterRef, Prefetch, Q, Subquery, Value
 
+from django_filters.rest_framework import DjangoFilterBackend
 from rest_framework import (
     decorators,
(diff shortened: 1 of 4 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -316,6 +317,15 @@ class TeamAccessViewSet(
     queryset = models.TeamAccess.objects.all().select_related("user")
     list_serializer_class = serializers.TeamAccessReadOnlySerializer
     detail_serializer_class = serializers.TeamAccessSerializer
+    filter_backends = (DjangoFilterBackend, filters.OrderingFilter)
+    ordering_fields = [
+        "role",
+        "user__email",
+        "user__name",
+        "user__main_identity__email",
+        "user__main_identity__name",
```
mjeammet, line 326: Syntax is correct but it doesn't seem to be working because it should refer a field.

PR #91, opened 2024-03-06
💫 Text errors duration

AntoLC:
## Purpose

Set duration on text errors coming from API.

## Proposal

- [x] Add effect prop on Box component
- [x] Set duration on text errors coming from API

## Demo

[scrnli_3_6_2024_11-05-40 AM.webm](https://github.com/numerique-gouv/people/assets/25994652/49ea1707-ad91-49c0-843b-094ef7a135f7)

### DIFF
--- a/src/frontend/apps/desk/src/components/Box.tsx
+++ b/src/frontend/apps/desk/src/components/Box.tsx
@@ -2,6 +2,8 @@ import { ComponentPropsWithRef, ReactHTML } from 'react';
 import styled from 'styled-components';
 import { CSSProperties } from 'styled-components/dist/types';
 
+import { hideEffect, showEffect } from './Effect';
+
 export interface BoxProps {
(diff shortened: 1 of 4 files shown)


lebaudantoine:
I haven't reviewed the code, but few comment on the feature itself.
I luv the text animation.

- Do you think it's user-friendly having the error disappears before the next form submission? 
- The form input should be in error mode (red outline) to make to follow Cunningham rules?

I guess we can ask Johann to help us make a decision on these interactivity features.

lebaudantoine (commented):
Nice addition, please read my comment on the feature itself from a UX perspective.

I'm also curious whether we should consider testing a shorter timeout in the end-to-end (e2e) test. It seems logical that if an error takes twice as long to appear than expected, the test should not pass. What are your thoughts on adjusting the timeout for a more stringent validation?

PR #96, opened 2024-03-06
Enhance user search

lebaudantoine:
## Purpose

We introduce user's name, we should be able to search user both on their email and name

## Proposal

Inspired by `Contact` search feature, refactor the current search logic on the user, to include a trigram similarity score on her name, summed up with the existing one, on her email.

Tests might be improved, wdyt @sampaccoud?

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -2,6 +2,7 @@
 
 from django.contrib.postgres.search import TrigramSimilarity
 from django.db.models import Func, Max, OuterRef, Prefetch, Q, Subquery, Value
+from django.db.models.functions import Coalesce
 
 from rest_framework import (
     decorators,
@@ -18,9 +19,7 @@ from core import models
 
 from . import permissions, serializers
 
-EMAIL_SIMILARITY_THRESHOLD = 0.01
-# TrigramSimilarity threshold is lower for searching email than for names,
-# to improve matching results
+SIMILARITY_THRESHOLD = 0.04
 
 
 class NestedGenericViewSet(viewsets.GenericViewSet):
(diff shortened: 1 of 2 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -214,10 +212,14 @@ def get_queryset(self):
                         Func("identities__email", function="unaccent"),
                         Func(Value(query), function="unaccent"),
                     )
+                    + TrigramSimilarity(
+                        Func("identities__name", function="unaccent"),
+                        Func(Value(query), function="unaccent"),
+                    )
```
lebaudantoine, line 218: I might need to add a bit more tests, it feels a bit random when searching user. I have decided to follow contact's implementation, but I could opt for another approach

sampaccoud:
@lebaudantoine Not sure what you mean by random results... It surely isn't random. Scores shouldn't be too hard to check on simple names/emails like you have in your tests. 
This being said, I would say open an issue and come back to it later when we have a bigger database in production because the edge cases will come to you and will be obvious. 

About tests, you have 3 types of tests:
- battle testing search on email : already done
- battle testing search on name
- testing interaction between name and email on a few edge cases?

lebaudantoine:
The order when summing the Trigram scores matters, and will change order of results, which I would have not expected.

lebaudantoine:
To summarize:
- I added a minimum test coverage to the changes introduced.
- I opened an issue, requesting advanced tests on this feature. #105 
- I have added a default empty string value on email and name when computing their trigram similarity score. I explain why in the commit. Please feel free to ask any question

We should be good to go, for this nice-to-have enhancement

src/backend/core/api/viewsets.py
```diff
@@ -2,6 +2,7 @@
 
 from django.contrib.postgres.search import TrigramSimilarity
 from django.db.models import Func, Max, OuterRef, Prefetch, Q, Subquery, Value
+from django.db.models.functions import Coalesce
```
sampaccoud, line 5: in commit message: 
s/Triagram/Trigram

sampaccoud: I didn't understand this statement in the commit message:
> Thus, the total score would reflect
the non-none similarity score.

sampaccoud (approved):
Nicely done with great explanations :pray:

src/backend/core/api/viewsets.py
lebaudantoine: thx for your feedback ! I have updated commit's body

PR #97, opened 2024-03-06
🩹(frontend) fix layout overflow in Team info

### LINKED ISSUE #84: Team Info Responsiveness
## Bug Report

**Problematic behavior**
Minor responsiveness issues on small screen.

`created_at` and `updated_at` data should be displayed inline. The text wrap without including the date, which hurts readability.

Capture d’écran 2024-03-04 à 11 32 07

On very small screens, the text overflow the card.
Capture d’écran 2024-03-04 à 11 32 21

lebaudantoine:
## Purpose

Enhance Team info layout

## Proposal

Few minor layout issues were fixed.

First display label and dates inline, so they wrap nicely when screen's size decreases. It also fixes the text overflow when the screen's size is tiny.

Then, align screen with the Figma design, where items are justified on the left of the Team info component.
It closes #84 

https://github.com/numerique-gouv/people/assets/45729124/3c745a97-9221-484b-8b3d-0e9e5e761ccf

### DIFF
--- a/src/frontend/apps/desk/src/features/teams/components/TeamInfo.tsx
+++ b/src/frontend/apps/desk/src/features/teams/components/TeamInfo.tsx
@@ -58,21 +58,30 @@ export const TeamInfo = ({ team }: TeamInfoProps) => {
       </Box>
       <Box
         className="p-s"
-        $gap="1rem"
+        $gap="2rem"
         $direction="row"
-        $justify="space-evenly"
-        $css={`border-top: 1px solid ${colorsTokens()['card-border']};`}
+        $justify="start"
+        $css={`
+        border-top: 1px solid ${colorsTokens()['card-border']};
+        padding-left: 1.5rem;
+        @media (min-width: 768px) {
+          padding-left: 6rem;
+        }`}
       >
-        <Text $size="s">
+        <Text $size="s" as="p">
           {t('{{count}} member', { count: team.accesses.length })}
         </Text>
-        <Text $size="s" $direction="row">
+        <Text $size="s" $display="inline" as="p">
           {t('Created at')}&nbsp;
-          <Text $weight="bold">{created_at}</Text>
+          <Text $weight="bold" $display="inline">
+            {created_at}
+          </Text>
         </Text>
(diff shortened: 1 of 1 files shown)


src/frontend/apps/desk/src/features/teams/components/TeamInfo.tsx
```diff
@@ -58,21 +58,25 @@ export const TeamInfo = ({ team }: TeamInfoProps) => {
...
+        $gap="2rem"
         $direction="row"
-        $justify="space-evenly"
+        $justify="start"
         $css={`border-top: 1px solid ${colorsTokens()['card-border']};`}
       >
-        <Text $size="s">
+        <Text $size="s" as="p">
           {t('{{count}} member', { count: team.accesses.length })}
         </Text>
```
lebaudantoine, line 68: In plural + In English, this element is not translated. Any clue @AntoLC ?

AntoLC (approved):
Try to fit a bit more with the mockup, `padding-left` or something, otherwise seems good. ✅ 

image

image

src/frontend/apps/desk/src/features/teams/components/TeamInfo.tsx
AntoLC: Ah yes I know why... it is because we don't translate in english, but actually we should in these cases. 
Even if the keys are in english they are still keys, to be corrected we must have a value bind to the key, in english we don't have. Don't know if it is clear, I will open a issue about it.

AntoLC: https://github.com/numerique-gouv/people/issues/99

lebaudantoine:
@AntoLC I have slightly improved the layout to make test aligned. I won't spend much time on it, however some nit-picking elements could be improved. I am a bit scared that text-alignments could be broken as soon as we update the image.

lebaudantoine:
https://github.com/numerique-gouv/people/assets/45729124/05438e24-4a20-4f7b-a0b0-34e8e46351ad

PR #98, opened 2024-03-07
🎨(backend) early return in User factory

lebaudantoine:
Avoid unnecessary nesting when code can early return.
Also, rename "item" to a more explicit name "user_entry".

Yesss this is very-very nit-picking 🥵


mjeammet (approved):
Let's go !

Issue #90, opened 2024-03-06
Enhance the make demo


lebaudantoine:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

The current implementation of our product demo via the make command lacks user identity for a significant portion of generated users, limiting the realism of the showcased scenarios. As it stands, users created by the make command lack complete information, such as full names and email addresses, because they don't have any identity.

**Describe the solution you'd like**

I propose enhancing the make command to create a more diverse set of user profiles for a more comprehensive product demonstration. The improvements could include:

- Making users without any identity, still present but relatively rare.
- Incorporating users with partial information in their identity, such as those with missing email addresses or full names.

This refinement in user generation will enable a more nuanced representation of user data scenarios, providing a more accurate and compelling demonstration of our product within various use cases.

PR #100, opened 2024-03-08
🧑‍💻(demo) improve distribution in number of identities per user

sampaccoud:
## Purpose

The current implementation of our product demo via the make command lacks user identity for a significant portion of generated users, limiting the realism of the showcased scenarios. As it stands, users created by the make command lack complete information, such as full names and email addresses, because they don't have any identity.

## Proposal 

I tried to come up with the simplest solution: we now generate a very small portion of our users with 0 identities. The probability for users to have only 1 identity is the highest but they can have up to 4 with decreasing probabilities. I removed the possibility to set a maximum number of identities as it doesn't bring any value.

3% of the identities created will have no email and 3% will have no name.

Fixes https://github.com/numerique-gouv/people/issues/90

### DIFF
--- a/src/backend/demo/management/commands/create_demo.py
+++ b/src/backend/demo/management/commands/create_demo.py
@@ -129,16 +129,19 @@ def create_demo(stdout):
         users_values = list(models.User.objects.values("id", "email"))
         for user_dict in users_values:
             for i in range(
-                random.randint(0, defaults.NB_OBJECTS["max_identities_per_user"])
+                random.choices(range(5), weights=[5, 50, 30, 10, 5], k=1)[0]
             ):
                 user_email = user_dict["email"]
                 queue.push(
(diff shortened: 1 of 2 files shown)


src/backend/demo/management/commands/create_demo.py
```diff
@@ -129,16 +129,19 @@ def create_demo(stdout):
...
             ):
                 user_email = user_dict["email"]
                 queue.push(
                     models.Identity(
                         user_id=user_dict["id"],
                         sub=uuid4(),
-                        email=f"identity{i:d}{user_email:s}",
                         is_main=(i == 0),
-                        name=fake.name(),
+                        # Leave 5% of emails and names empty
```
lebaudantoine, line 140: *`3%`

lebaudantoine (approved):

PR #101, opened 2024-03-08
🌐(app-desk) translate role in member grid

AntoLC:
## Purpose

The roles in the member grid were not being translated. This PR adds the translation for
the roles in the member grid.

## Proposal

- [x] Add translated roles

## Demo

image

### DIFF
--- a/src/frontend/apps/desk/src/features/teams/components/Member/MemberGrid.tsx
+++ b/src/frontend/apps/desk/src/features/teams/components/Member/MemberGrid.tsx
@@ -33,6 +33,12 @@ export const MemberGrid = ({ teamId, currentRole }: MemberGridProps) => {
     setPagesCount(data?.count ? Math.ceil(data.count / pageSize) : 0);
   }, [data?.count, pageSize, setPagesCount]);
 
+  const dictRole = {
+    [Role.ADMIN]: t('Admin'),
+    [Role.MEMBER]: t('Member'),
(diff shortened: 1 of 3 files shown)


src/frontend/apps/desk/src/features/teams/components/Member/MemberGrid.tsx
```diff
@@ -78,8 +84,11 @@ export const MemberGrid = ({ teamId, currentRole }: MemberGridProps) => {
             headerName: t('Emails'),
           },
           {
-            field: 'role',
+            id: 'role',
```
lebaudantoine, line 87: Reading this update, I wonder if ID on custom cell should be set to a unique value per row. Same question applies for the avatar column. A unique ID is a general good practice to avoid unnecessary re-render, wdyt? I read Cunningham documentation, and found no more info about it.
Let's keep it as it is. Nice one !!

lebaudantoine (approved):

src/frontend/apps/desk/src/features/teams/components/Member/MemberGrid.tsx
AntoLC: The `id` seems to be only for the column, so it should be ok:
https://github.com/openfun/cunningham/blob/30ec253ab0b7f655f90152f3f0797a8493f0ee2b/packages/react/src/components/DataGrid/utils.tsx#L27

PR #109, opened 2024-03-12
Create a Docker Image for Frontend sources

lebaudantoine:
## Purpose

Prepare project deployment.

## Proposal

Since DevOps lacks a certified cold storage solution (e.g., S3) for serving static files, we've decided to containerize the frontend as a quick workaround for deploying staging environments.

I tried to explain precisely my choices, or the current WIP state of each element in my commits. Please feel free to ask any question.


Dockerfile
```diff
@@ -11,6 +11,40 @@ RUN apt-get update && \
...
   rm -rf /var/lib/apt/lists/*
 
+
+### ---- Front-end builder image ----
+FROM node:20 as front-builder
+
+# Copy frontend app sources
+COPY ./src/frontend /builder
+
+WORKDIR /builder/apps/desk
```
lebaudantoine, line 21: I should copy only desk

lebaudantoine: based on @AntoLC use smth similar as https://github.com/openfun/marsha/blob/129de945a5fada632ba3faf6538a6a19525b7343/Dockerfile#L36
to cache dependencies

Dockerfile
```diff
@@ -11,6 +11,40 @@ RUN apt-get update && \
...
+
+# Un-privileged user running the application
+ARG DOCKER_USER
+USER ${DOCKER_USER}
+
+COPY --from=front-builder \
+    /builder/apps/desk/out \
+    /usr/share/nginx/html
+
+COPY ./src/frontend/apps/desk/conf/default.conf /etc/nginx/conf.d:ro
```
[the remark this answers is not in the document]
lebaudantoine: good catch

Dockerfile
lebaudantoine: Alright, I took into account your feedbacks, and slightly modified your suggestion, but I am not 100% sure if I should copy all package.json before running `yarn install`. I red again the Marsha docker file, and they seem to follow the latter approach. What do you think?

lebaudantoine: I refacto this part, lmk if it's better!

Dockerfile
```diff
@@ -11,6 +11,53 @@ RUN apt-get update && \
...
+COPY ./src/frontend/packages/eslint-config-people/package.json ./packages/eslint-config-people/package.json
+
+RUN yarn --frozen-lockfile
+
+### ---- Front-end builder image ----
+FROM node:20 as frontend-builder
+
+WORKDIR /app
+
+COPY --from=frontend-deps /app/node_modules ./node_modules
```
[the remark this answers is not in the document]
lebaudantoine: I took inspiration from Nextjs docker image

PR #114, opened 2024-03-15
✨(backend) convert invitations to accesses

mjeammet:
## Purpose

Convert invitations to accesses upon creating a new identity.

## Proposal

- [x] modify the `save` method on Identity (could have been in pre or post_save signals but Django recommends against it)
- [x] write a positive test : invitation converted and access granted
- [x] write a negative test for expired invitations

I don't remember why but I had noted down that we could add the logic of this PR in `create_user` ([here](https://github.com/numerique-gouv/people/blob/29d274ab7c94768f30752319cabbcc1aee7db142/src/backend/core/authentication.py#L108))
I think it's better to link it to the Identity model because identities and invitations both revolve around the same emails. We could imagine by instance that, later on, upon linking identities to my existing user, I get all the extra accesses linked to my new email/identity.

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -284,8 +284,34 @@ class Identity(BaseModel):
         return f"{id_str:s}{main_str:s}"
 
     def save(self, *args, **kwargs):
-        """Ensure users always have one and only one main identity."""
+        """
+        Saves identity, ensuring users always have exactly one main identity.
+        Also converts valid invitations to accesses upon creating an identity.
+        """
+
+        # If new identity, convert all valid invitations to team accesses.
+        # Expired invitations are ignored.
+        if self._state.adding:
+            active_invitations = Invitation.objects.filter(
+                email=self.email,
+                created_at__gte=timezone.now()
+                - timedelta(seconds=settings.INVITATION_VALIDITY_DURATION),
+            ).select_related("team")
+
+            if active_invitations.exists():
+                TeamAccess.objects.bulk_create(
+                    [
+                        TeamAccess(
+                            user=self.user, team=invitation.team, role=invitation.role
+                        )
+                        for invitation in active_invitations
+                    ]
+                )
+                active_invitations.delete()
+
         super().save(*args, **kwargs)
+
+        # Ensure users always have one and only one main identity.
         if self.is_main is True:
             self.user.identities.exclude(id=self.id).update(is_main=False)
 
--- a/src/backend/core/tests/test_models_invitations.py
+++ b/src/backend/core/tests/test_models_invitations.py
@@ -3,15 +3,21 @@ Unit tests for the Invitation model
 """
 
 import time
+import uuid
 
 from django.contrib.auth.models import AnonymousUser
 from django.core import exceptions
 
 import pytest
+from faker import Faker
+from freezegun import freeze_time
 
-from core import factories
+from core import factories, models
 
 pytestmark = pytest.mark.django_db
+
+
+fake = Faker()
 
 
 def test_models_invitations_readonly_after_create():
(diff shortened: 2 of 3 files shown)


src/backend/core/models.py
```diff
@@ -284,8 +284,27 @@ def __str__(self):
...
+        if self._state.adding is True:
+            validity_duration = timedelta(seconds=settings.INVITATION_VALIDITY_DURATION)
+            active_invitations = Invitation.objects.filter(
+                email=self.email, created_at__gte=timezone.now() - validity_duration
+            )
+            for invitation in active_invitations:
+                TeamAccess.objects.create(
+                    user=self.user, team=invitation.team, role=invitation.role
+                )
+                invitation.delete()
```
lebaudantoine, line 303: Have you though using a `bulk_create()`? [doc](https://docs.djangoproject.com/en/4.1/ref/models/querysets/#bulk-create).

It would save us at least n queries. However, I think bulk delete doesn't exist, we might still need to call it n-times.
Wdyt?

src/backend/core/models.py
```diff
@@ -284,8 +284,27 @@ def __str__(self):
         return f"{id_str:s}{main_str:s}"
 
     def save(self, *args, **kwargs):
-        """Ensure users always have one and only one main identity."""
+        """
+        Save identity, ensuring user always have one and only one main identity.
+        Also convert valid invitations if new identity.
```
lebaudantoine, line 289: might be cool to explain to what are converted the valid invitation !

src/backend/core/models.py
```diff
@@ -284,8 +284,27 @@ def __str__(self):
...
+        """
+
+        # If new identity, convert all valid invitations to team accesses.
+        # Expired invitations are ignored.
+        if self._state.adding is True:
+            validity_duration = timedelta(seconds=settings.INVITATION_VALIDITY_DURATION)
+            active_invitations = Invitation.objects.filter(
+                email=self.email, created_at__gte=timezone.now() - validity_duration
+            )
```
lebaudantoine, line 298: I know `@property` are not accessible when using filter, however this feel duplicating a critical logic. I have better solution in mind, I just wanted to discuss it, anyway feel free to ingore my comment

src/backend/core/tests/test_models_invitations.py
```diff
@@ -73,6 +73,35 @@ def test_models_invitations__is_expired(settings):
...
 
 
+def test_models_invitation__new_user__convert_invitation_to_access():
+    """
+    Upon creating a new identity, invitations linked to that email
+    should be converted to accesses and then deleted.
+    """
+    invitation = factories.InvitationFactory()
+    new_identity = factories.IdentityFactory(is_main=True, email=invitation.email)
+
```
lebaudantoine, line 83: I would create some other invitation to make sure only the one with a matching email is converted

src/backend/core/tests/test_models_invitations.py
```diff
@@ -73,6 +73,35 @@ def test_models_invitations__is_expired(settings):
...
+    assert models.Invitation.objects.exists() is False
+    assert models.TeamAccess.objects.exists() is True
```
lebaudantoine, line 85: Same here, I would be more specific when checking the created TeamAccess, wdyt? I would check the new object in db instead of the abilities, but I might wrong

src/backend/core/tests/test_models_invitations.py
```diff
@@ -73,6 +73,35 @@ def test_models_invitations__is_expired(settings):
...
+
+
+def test_models_invitation__new_user__expired_invitation(settings):
+    """
+    Upon creating a new identity, expired invitations
+    should not be converted into accesses.
+    """
+    invitation = factories.InvitationFactory()
+    settings.INVITATION_VALIDITY_DURATION = 1
+    time.sleep(1)
```
lebaudantoine, line 96: cool, using freeze gun could help us testing the default validity duration

src/backend/core/tests/test_models_invitations.py
```diff
@@ -73,6 +73,35 @@ def test_models_invitations__is_expired(settings):
...
+
+    new_identity = factories.IdentityFactory(is_main=True, email=invitation.email)
+
+    assert models.Invitation.objects.exists() is True
+    assert models.TeamAccess.objects.exists() is False
+    assert invitation.team.get_abilities(new_identity.user)["get"] is False
```
lebaudantoine, line 102: same comment

lebaudantoine (commented):
Few comments, please feel free to discuss it in MP

src/backend/core/models.py
mjeammet: I hadn't ! I've looked into "bulk_delete" solutions but none of them seem satisfying to me. 

Would you use `bulk_create` if we still have to use a for loop for deletion ?

src/backend/core/models.py
mjeammet: it's explained just below but I'll add it to the docstrings too.

src/backend/core/models.py
lebaudantoine: do you have any input @sampaccoud?

src/backend/core/tests/test_models_invitations.py
mjeammet: I still think it's a bit overkill for unit tests such as these.

src/backend/core/tests/test_models_invitations.py
mjeammet: I modified it slightly to be more specific.

src/backend/core/tests/test_models_invitations.py
```diff
@@ -73,6 +73,51 @@ def test_models_invitations__is_expired(settings):
...
+            team=invitation.team, user=new_identity.user
+        ).exists()
+        is False
+    )
+    assert (
+        models.Invitation.objects.filter(
+            team=invitation.team, email=invitation.email
+        ).exists()
+        is True
+    )
```
lebaudantoine, line 118: nit: 
```suggestion
    assert not  models.TeamAccess.objects.filter(team=invitation.team, user=new_identity.user).exists()
    assert models.Invitation.objects.filter(team=invitation.team, email=invitation.email).exists()
```
should do the same, I let you lint this suggestion

src/backend/core/tests/test_models_invitations.py
```diff
@@ -73,6 +73,51 @@ def test_models_invitations__is_expired(settings):
...
+
+    assert (
+        models.TeamAccess.objects.filter(
+            team=invitation.team, user=new_identity.user
+        ).exists()
+        is True
+    )
+    assert (
+        len(models.Invitation.objects.filter(team=invitation.team)) == 1
+    )  # the other invitation remains
```
lebaudantoine, line 93: This test could slightly be improved by checking that the right invitations were deleted.

src/backend/core/tests/test_models_invitations.py
```diff
@@ -73,6 +73,51 @@ def test_models_invitations__is_expired(settings):
     assert expired_invitation.is_expired is True
 
 
+def test_models_invitation__new_user__convert_invitation_to_access():
```
lebaudantoine, line 76: We should test :
- One invitation to xyw@email.com converted to a team access ✅
- Several invitations to xyw@email.com converted to several team accesses
- Expired invitations kept in the database with no team access created ✅

Moreover, we test in isolation valid and expired invitations, we have no test to check that only valid invitations will be converted when an email has several invitations with some expired.

src/backend/core/models.py
```diff
@@ -284,8 +284,31 @@ def __str__(self):
...
+            TeamAccess.objects.bulk_create(
+                [
+                    TeamAccess(
+                        user=self.user, team=invitation.team, role=invitation.role
+                    )
+                    for invitation in active_invitations
+                ]
+            )
+            active_invitations.delete()
```
lebaudantoine, line 307: niceee, have you inspected the number of SQL queries, to make sure only two are triggered now?

mjeammet: Unfortunately, django_assert_num_queries reports 16 queries + 1 for each invitation :/  

`.bulk_create()` and `.delete()` are only one query each, so that's nice.
But the comprehensive list inside the `bulk_create()` results in this query being repeated for each invitation. I'm not sure I understand.

```
SELECT "people_team"."id", "people_team"."created_at", "people_team"."updated_at", "people_team"."name", "people_team"."slug" FROM "people_team" WHERE "people_team"."id" = '2c5e952e221f47d59cfadf079295c56a'::uuid LIMIT 21
```

mjeammet: oh my god, is this a select_related situation ? :o

mjeammet: OMG, it was. 💡🧠💡 !!!

src/backend/core/models.py
```diff
@@ -284,8 +284,31 @@ def __str__(self):
...
+        Saves identity, ensuring users always have exactly one main identity.
+        Also converts valid invitations to accesses upon creating an identity.
+        """
+
+        # If new identity, convert all valid invitations to team accesses.
+        # Expired invitations are ignored.
+        if self._state.adding:
+            validity_duration = timedelta(seconds=settings.INVITATION_VALIDITY_DURATION)
+            active_invitations = Invitation.objects.filter(
+                email=self.email, created_at__gte=timezone.now() - validity_duration
```
lebaudantoine, line 297:
```suggestion
            active_invitations = Invitation.objects.filter(
                email=self.email, created_at__gte=timezone.now() - timedelta(seconds=settings.INVITATION_VALIDITY_DURATION)
```

src/backend/core/models.py
lebaudantoine: nice investigation 🕵️ I would test it

src/backend/core/tests/test_models_invitations.py
```diff
@@ -73,6 +79,97 @@ def test_models_invitations__is_expired(settings):
...
+):
+    """
+    The number of queries executed during user creation should not be proportional
+    to the number of invitations being processed.
+    """
+    user_email = fake.email()
+
+    if num_invitations != 0:
+        for _ in range(0, num_invitations):
+            factories.InvitationFactory(email=user_email, team=factories.TeamFactory())
```
lebaudantoine, line 158: nit: `create_batch()` could do the magic 😍

lebaudantoine (approved):
LGTM

src/backend/core/tests/test_models_invitations.py
mjeammet: I've tried but factories.TeamFactory() creates one factory and not many. It ends up in django throwing a unique constraints validation error.

Issue #81, opened 2024-03-03
Add ordering on Team Accesses


lebaudantoine:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Based on recent design changes, we need to support ordering on any column of the data grid.

<img width="1140" alt="Capture d’écran 2024-03-03 à 23 19 28" src="https://github.com/numerique-gouv/people/assets/45729124/8013e210-3831-46af-a499-85d613df513d">

**Describe the solution you'd like**
Refactor the `Pagination` class, to support more than one ordering method, actually it's limited to the `created_on` field.

lebaudantoine:
@AntoLC you will need this feature to support column's ordering on the data grid.

lebaudantoine:
I am working on it

PR #115, opened 2024-03-15
✨(api) support ordering when listing team's accesses

lebaudantoine:
## Purpose

Offer the necessary ordering field to enhance UX when manipulating Team's data in the member grid.

## Proposal

Add API ordering fields on the TeamAccess endpoint. Please note that, frontend is almost ready, and being blocked by issue on Cunnigham side. One was merged, another is still being discussed. 

I prefer shipping the backend first, and opening a new PR for the frontend when all elements will be available.

This PR is related to #81

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -327,8 +327,8 @@ class TeamAccessViewSet(
     detail_serializer_class = serializers.TeamAccessSerializer
 
     filter_backends = [filters.OrderingFilter]
-    ordering = ['role']
-    ordering_fields = ['role']
+    ordering = ["role"]
+    ordering_fields = ["role", "email", "name"]
 
     def get_permissions(self):
         """User only needs to be authenticated to list team accesses"""
@@ -361,6 +361,10 @@ class TeamAccessViewSet(
                 user=self.request.user, team=self.kwargs["team_id"]
             ).values("role")[:1]
 
+            user_main_identity_query = models.Identity.objects.filter(
+                user=OuterRef("user_id"), is_main=True
+            )
+
             queryset = (
                 # The logged-in user should be part of a team to see its accesses
                 queryset.filter(
@@ -375,7 +379,11 @@ class TeamAccessViewSet(
                 )
                 # Abilities are computed based on logged-in user's role and
                 # the user role on each team access
-                .annotate(user_role=Subquery(user_role_query))
+                .annotate(
+                    user_role=Subquery(user_role_query),
+                    email=Subquery(user_main_identity_query.values("email")[:1]),
+                    name=Subquery(user_main_identity_query.values("name")[:1]),
+                )
                 .distinct()
             )
         return queryset
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -326,6 +326,10 @@ class TeamAccessViewSet(
     list_serializer_class = serializers.TeamAccessReadOnlySerializer
     detail_serializer_class = serializers.TeamAccessSerializer
 
+    filter_backends = [filters.OrderingFilter]
+    ordering = ['role']
+    ordering_fields = ['role']
+
     def get_permissions(self):
         """User only needs to be authenticated to list team accesses"""
         if self.action == "list":
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -97,7 +97,6 @@ class SerializerPerActionMixin:
 class Pagination(pagination.PageNumberPagination):
     """Pagination to display no more than 100 objects per page sorted by creation date."""
 
-    ordering = "-created_on"
     max_page_size = 100
     page_size_query_param = "page_size"
 
--- a/src/backend/core/factories.py
+++ b/src/backend/core/factories.py
@@ -124,6 +124,7 @@ class UserFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = models.User
+        django_get_or_create = ("email",)
 
     email = factory.Faker("email")
     language = factory.fuzzy.FuzzyChoice([lang[0] for lang in settings.LANGUAGES])
--- a/src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
+++ b/src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
@@ -203,3 +203,45 @@ def test_api_team_accesses_list_authenticated_constant_numqueries(
 
     assert response.status_code == 200
     assert response.json()["count"] == 21
+
+
+def test_api_team_accesses_list_authenticated_ordering():
+    """Team accesses can be ordered by "role"."""
+
+    user = factories.UserFactory()
+    factories.IdentityFactory(user=user, is_main=True)
+
+    team = factories.TeamFactory()
+    models.TeamAccess.objects.create(team=team, user=user)
+
+    # create 20 new team members
+    for _ in range(20):
+        extra_user = factories.IdentityFactory(is_main=True).user
+        factories.TeamAccessFactory(team=team, user=extra_user)
+
+    client = APIClient()
+    client.force_login(user)
+
+    response = client.get(
+        f"/api/v1.0/teams/{team.id!s}/accesses/?ordering=role",
+    )
+    assert response.status_code == 200
+    assert response.json()["count"] == 21
+
+    results = [
+        team_access["role"]
+        for team_access in response.json()["results"]
+    ]
+    assert sorted(results) == results
+
+    response = client.get(
+        f"/api/v1.0/teams/{team.id!s}/accesses/?ordering=-role",
+    )
+    assert response.status_code == 200
+    assert response.json()["count"] == 21
+
+    results = [
+        team_access["role"]
+        for team_access in response.json()["results"]
+    ]
+    assert sorted(results, reverse=True) == results
--- a/src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
+++ b/src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
@@ -3,6 +3,7 @@ Test for team accesses API endpoints in People's core app : list
 """
 
 import pytest
+from rest_framework.status import HTTP_200_OK
 from rest_framework.test import APIClient
 
 from core import factories, models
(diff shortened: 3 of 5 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -373,6 +380,8 @@ def get_queryset(self):
                 # Abilities are computed based on logged-in user's role and
                 # the user role on each team access
                 .annotate(user_role=Subquery(user_role_query))
+                .annotate(email=Subquery(user_main_identity.values("email")[:1]))
+                .annotate(name=Subquery(user_main_identity.values("name")[:1]))
```
lebaudantoine, line 384: I could have done also smth as :
```
    .annotate(
        email=Max(
            "user__identities__email",
            filter=Q(user__identities__is_main=True),
        )
    )
    .annotate(
        name=Max(
            "user__identities__name",
            filter=Q(user__identities__is_main=True),
        )
    )

```
lmk what do you prefer

src/backend/core/tests/test_api_users.py
```diff
@@ -440,6 +440,56 @@ def test_api_users_list_pagination(
...
+
+    response = client.get(
+        f"/api/v1.0/users/?page_size={page_size}",
+    )
+
+    assert response.status_code == HTTP_200_OK
+    content = response.json()
+
+    assert content["count"] == page_size + 1
+    assert len(content["results"]) == page_size
```
mjeammet, line 464: I feel like you could have used the test just above, WDYT ?

lebaudantoine: Actually maybe, but test just above mock page_size, so how could set it through a query parameter? The test just above actually tests the pagination mechanism, more than the page size. I preferred having simpler tests, that tests a well-scoped unit of code. However that's a personal opinion!

src/backend/core/api/viewsets.py
lebaudantoine: Also, please let consider a future refactoring. We are joining twice the user-related data, using both annotate and prefetch. Could we do a simple join?

mjeammet: This is what is was thinking about the other day : [order_by at the same moment than the prefetch_related](https://stackoverflow.com/questions/48071397/django-prefetch-relateds-prefetch-order-by). I'm not sure if we can use it to avoid another annotate or query but I wanted to share it to you.

mjeammet: > I could have done also smth as :
> 
> ```
>     .annotate(
>         email=Max(
>             "user__identities__email",
>             filter=Q(user__identities__is_main=True),
>         )
>     )
>     .annotate(
>         name=Max(
>             "user__identities__name",
>             filter=Q(user__identities__is_main=True),
>         )
>     )
> ```
> 
> lmk what do you prefer

That's nice too but the "Max" kinda confuses me. Does that remove 2 queries from subqueries, though ? or are they already "included" in the "main" query ?

src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
```diff
@@ -203,3 +203,82 @@ def test_api_team_accesses_list_authenticated_constant_numqueries(
...
+    client = APIClient()
+    client.force_login(user)
+
+    response = client.get(
+        f"/api/v1.0/teams/{team.id!s}/accesses/?ordering=role",
+    )
+    assert response.status_code == 200
+    assert response.json()["count"] == 21
+
+    results = [team_access["role"] for team_access in response.json()["results"]]
```
mjeammet, line 231: why use roles ?

src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
```diff
@@ -203,3 +203,82 @@ def test_api_team_accesses_list_authenticated_constant_numqueries(
...
+    assert sorted(results, reverse=True) == results
+
+
+@pytest.mark.parametrize("ordering_fields", ["name", "email"])
+def test_api_team_accesses_list_authenticated_ordering_user(ordering_fields):
```
mjeammet, line 245: I feel like there's a smart way to factorize those two tests of same purpose using 

```
@pytest.mark.parametrize(
    "ordering_fields, method", 
    [
        ("role", "role"), 
        ("name", '["user"]["name"]'), 
        # ("email", "user__email")
    ])
```
It does raise a key error because of the presence absence of `["user"]` but I feel like a `if` just before the results would allow for a nice factorisation.

src/backend/core/api/viewsets.py
lebaudantoine: > This is what is was thinking about the other day : [order_by while prefetching](https://stackoverflow.com/questions/48071397/django-prefetch-relateds-prefetch-order-by). I'm not sure if we can use it to avoid another annotate or query but I wanted to share it to you.

Let's discuss it orally, I feel the order_by is applied on the second query made to prefect data.

lebaudantoine: Yeah, the Max is quite confusing. Actually the Subqueries don't add any query! I tried this approach, because I felt the SQL query under the hood was easier to understand. Let's stick to my initial approach

src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
```diff
@@ -203,3 +203,82 @@ def test_api_team_accesses_list_authenticated_constant_numqueries(
...
+        extra_user = factories.IdentityFactory(is_main=True).user
+        factories.TeamAccessFactory(team=team, user=extra_user)
+
+    client = APIClient()
+    client.force_login(user)
+
+    response = client.get(
+        f"/api/v1.0/teams/{team.id!s}/accesses/?ordering=role",
+    )
+    assert response.status_code == 200
```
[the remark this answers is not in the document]
lebaudantoine: Afterthought, and @sampaccoud comment on one of your PR, we might prefer constant value. I'll still fix up my PR and then refactor all constant values.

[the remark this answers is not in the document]
lebaudantoine: Actually, we are not using DRF's status codes in this file. Afterthough, I reconsidered Sam's comment on "you should not use constant value in your test" I think we should replace DRFS's status code constants by hardcoded value to keep our tests dumb as possible.

src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
lebaudantoine: Actually, I prefer having dumb test, instead of complicated tests parametrized and with if that change their logic. In this situation I understand your point, if you really need this factorization I'll do it

lebaudantoine: We should not chase test factorization

src/backend/core/api/viewsets.py
sampaccoud: +1 on @mjeammet's opinion that Max is confusing.

sampaccoud, line 384: Maybe simpler with
```suggestion
                .annotate(
                    user_role=Subquery(user_role_query),
                    email=Subquery(user_main_identity.values("email")[:1]),
                    name=Subquery(user_main_identity.values("name")[:1]),
                )
```

src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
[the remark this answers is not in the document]
sampaccoud: You know my position but the other day I also validated the commit proposing to add constants :laughing:  so... :man_shrugging: should we call it nitpicking?

src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
sampaccoud: +1 on the opinion that this is too smart and makes it much harder to understand the test and each of the 2 contexts you are trying to test because I should understand the "why" by reading the docstring.

src/backend/core/tests/test_api_users.py
sampaccoud: +1 on dedicated tests with a single goal.

src/backend/core/api/viewsets.py
lebaudantoine: fixed

src/backend/core/api/viewsets.py
```diff
@@ -358,6 +361,10 @@ def get_queryset(self):
                 user=self.request.user, team=self.kwargs["team_id"]
             ).values("role")[:1]
 
+            user_main_identity = models.Identity.objects.filter(
```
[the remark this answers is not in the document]
lebaudantoine: fixed to `user_main_identity_query` to be aligned with `user_role_query`

src/backend/core/api/viewsets.py
```diff
@@ -327,6 +326,10 @@ class TeamAccessViewSet(
     list_serializer_class = serializers.TeamAccessReadOnlySerializer
     detail_serializer_class = serializers.TeamAccessSerializer
 
+    filter_backends = [filters.OrderingFilter]
+    ordering = ["role"]
+    ordering_fields = ["role", "email", "name"]
```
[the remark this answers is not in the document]
lebaudantoine: Not fan, reading all sources from Joanie and other openfun projects, I did not come across such comment.

src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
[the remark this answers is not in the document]
lebaudantoine: It might .... but I removed all these hard-coded values in my latest commit. I explain why in the commit's body. I think we took the wrong decision, I made a mistake, and I prefer fixing it rn. nothing that important btw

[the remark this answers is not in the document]
mjeammet: I still think status codes are clearer. Reading the name and not having to remember what this number means alleviates our brain charges, if even of 1%. It doesn't matter most of the time but sometimes it saves us (or at least me) some time.

Should almost count for accessibility for ADHD, I guess.

src/backend/core/factories.py
```diff
@@ -124,6 +124,7 @@ class UserFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = models.User
+        django_get_or_create = ("email",)
```
mjeammet, line 127: nit : did it not get or create before ? 🤔

src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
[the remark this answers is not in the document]
mjeammet: Regarding your comments on the commit body : 
- I don't think people used to the codes are actually disturbed by a constant being "<the code>_<the explanation"
- as for freezegun, we add the necessary packages if we decide they bring something to our test suite. Additionnaly, DRF is a dependency we already have.

src/backend/core/factories.py
lebaudantoine: I am not sure to understand, I aligned the `UserFactory` with the `IdentityFactory`

src/backend/core/tests/test_models_users.py
```diff
@@ -36,13 +36,13 @@ def test_models_users_email_unique():
...
     ):
-        factories.UserFactory(email=user.email)
+        models.User.objects.create(email=user.email)
 
 
 def test_models_users_email_several_null():
     """Several users with a null value for the "email" field can co-exist."""
     factories.UserFactory(email=None)
-    factories.UserFactory(email=None)
+    models.User.objects.create(email=None, password="foo.")
```
mjeammet, line 45: Nice to test both methods !

lebaudantoine: it's mandatory to make the test pass. If I am user the factory again, it won't create a new instance

src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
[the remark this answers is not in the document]
lebaudantoine: That's not the point, people used to use literals, will continue to use them. When reviewing their code, we'll nitpick those literals, they will have to fix up their code, which is totally functional, the reviewer will have to review again the code, and everyone will lose time for no added value. The code was functional with a literal, it will be with the constant.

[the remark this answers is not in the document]
lebaudantoine: When something doesn't matter, most of the time, we should focus on what's bring value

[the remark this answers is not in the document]
lebaudantoine: freezgun is a bit different, we need it to run/write our tests without create a monstrosity of flakiness. It will also improve code readability, having a context stating the date the object is created, is much more explicit than a sleep(1)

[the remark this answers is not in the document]
lebaudantoine: I will revert my commit and align my code to maintain the code accessible

src/backend/core/factories.py
mjeammet: I thought it already "get_or_created" the objet, but I imagine it only "created" it without being able to retrieve existing object.

mjeammet (approved):
WP 👏🏽

PR #130, opened 2024-03-19
🔧(backend) configure RedisCache in production settings

lebaudantoine:
## Purpose

Prepare staging deployment

## Description

Set-up Redis cache.

### DIFF
--- a/src/backend/people/settings.py
+++ b/src/backend/people/settings.py
@@ -291,6 +291,7 @@ class Base(Configuration):
 
     # Session
     SESSION_ENGINE = "django.contrib.sessions.backends.cache"
+    SESSION_CACHE_ALIAS = "default"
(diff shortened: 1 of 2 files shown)


src/backend/people/settings.py
```diff
@@ -291,6 +291,7 @@ class Base(Configuration):
 
     # Session
     SESSION_ENGINE = "django.contrib.sessions.backends.cache"
+    SESSION_CACHE_ALIAS = "default"
```
sampaccoud, line 294: I don't think you should use the default cache for this because people will use the default cache for something else. Call it "session"?

sampaccoud (approved):

src/backend/people/settings.py
lebaudantoine: fixed

lebaudantoine:
`SESSION_CACHE_ALIAS = "session"` make the tests fail. I will merge it with its "default" value, and look at it later on

PR #133, opened 2024-03-19
🐛(docker) switch CMD form from Shell to Exec

lebaudantoine:
## Purpose

Prepare deployment

## Proposal

`backend-development` and `backend-production` CMD syntaxes were using a Shell Form. Shell form prevented Unix signals from reaching our container correctly, such as SIGTERM. Also, the shell process ends up being the PID 1, instead of our Python scripts.

Docker recommends using the exec form whenever possible.


rouja (approved):
LGTM

PR #138, opened 2024-03-21
🎨(models) extract invitation converter in a proper method

lebaudantoine:
Improved code readability, by extracting this well-scoped unit of logic in a dedicated method. Also, rename active_invitations to match 'valid' vocabulary used elsewhere in the doc. If no valid invitation exists, early return to avoid nesting.

Also, I wonder @mjeammet should we add try/except to catch any database error, log them but ignore them when creating a new identity?


src/backend/core/models.py
```diff
@@ -286,35 +286,43 @@ def __str__(self):
     def save(self, *args, **kwargs):
         """
         Saves identity, ensuring users always have exactly one main identity.
-        Also converts valid invitations to accesses upon creating an identity.
+        If it's a new identity, give it access to the relevant teams.
```
mjeammet, line 289:
```suggestion
        If it's a new identity, give its user access to the relevant teams.
```

mjeammet:
> Improved code readability, by extracting this well-scoped unit of logic in a dedicated method. Also, rename active_invitations to match 'valid' vocabulary used elsewhere in the doc. If no valid invitation exists, early return to avoid nesting.

Absolute +1, both for renaming the invitations and the dedicated private method. 

> Also, I wonder @mjeammet should we add try/except to catch any database error, log them but ignore them when creating a new identity?

Good idea !

mjeammet (approved):

Issue #106, opened 2024-03-11
Enhance `user` string representation


lebaudantoine:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

For now, on `User` model, the `__str__` returns user's contact email first. It's actually something we don't use for now.

```
    def __str__(self):
        return (
            str(self.profile_contact)
            if self.profile_contact
            else self.email or str(self.id)
        )
```

**Describe the solution you'd like**
I would prefer that the string representation of the user mentions her main identity information (email and full name). It might make debugging easier.

jblagadec:
Hello
Can this ticket be moved to En Prod? (no testing for me)

PR #143, opened 2024-03-22
✨(webhook) add webhook logic and synchronization utils

mjeammet:
## Purpose

This webhook will allow us to send teams data to outside service providers. 

This feature is inspired by [webhooks in joanie](https://github.com/openfun/joanie/blob/ea1857dd33e79bb34d56b93fcb81b92303886b42/src/backend/joanie/core/utils/webhooks.py#L31
), which synchronize data in a similar way.

## Proposal

We write a utils method that take serialized team memberships and send them to an outline endpoint fit to receive and process the data.

- [x] calling scim synchronizer method upon adding or removing team accesses
- [x] write tests

I suggest we keep this PR limited to the synchronisation upon changes.
If necessary, another PR may follow, adding management command to trigger synchronisation in bash (for cron tasks by instance)


src/backend/core/tests/test_models_users.py
mjeammet, line 1: should we add a test for trying to access user name when no identity is provided ?

src/backend/core/models.py
```diff
@@ -471,6 +545,28 @@ def get_abilities(self, user):
...
+        choices=WebhookStatusChoices.choices,
+    )
+
+    class Meta:
+        db_table = "people_team_webhook"
+        verbose_name = _("Team webhook")
+        verbose_name_plural = _("Team webhooks")
+
+    def __str__(self):
+        return f"Webhook to {self.url} for {self.team}"
```
mjeammet, line 567: I didn't expect that ! You plan on having a webhook per team while I thought we had a webhook per application.

I see the value of having a status per team.

But if we plan on starting with Outline :
- there's no point in having a secret field, at least for Outline. As only administrators can change groups composition, I don't think we'll be using a different secret for each team (or we'll have to generate them programmatically and we'll have to check it is possible).
- for Outline again, instead of storing an URL, we should store UUIDs directly (to call API and infer URLs from it)

src/backend/core/migrations/0002_remove_user_email_user_admin_email_and_more.py
mjeammet, line 1: We had decided to have a single migration file until first release. OK to replace 0001 initial instead ?

src/backend/core/migrations/0002_remove_user_email_user_admin_email_and_more.py
```diff
@@ -0,0 +1,46 @@
...
+import django.db.models.deletion
+import uuid
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0001_initial'),
+    ]
```
mjeammet, line 12: We had decided to have a single migration file until first release. OK to replace 0001 initial instead ?

src/backend/core/tests/test_models_users.py
lebaudantoine: agree

src/backend/core/api/viewsets.py
```diff
@@ -363,13 +354,6 @@ def get_queryset(self):
...
-                .prefetch_related(
-                    Prefetch(
-                        "user__identities",
-                        queryset=models.Identity.objects.filter(is_main=True),
-                        to_attr="main_identity",
-                    )
-                )
                 # Abilities are computed based on logged-in user's role and
                 # the user role on each team access
                 .annotate(user_role=Subquery(user_role_query))
```
lebaudantoine, line 359: Fyi, when rebasing with main, you will pull my recent changes on this precise queryset, where I added the ordering.

src/backend/core/api/viewsets.py
```diff
@@ -203,7 +203,7 @@ def get_queryset(self):
                 Prefetch(
                     "identities",
                     queryset=models.Identity.objects.filter(is_main=True),
-                    to_attr="main_identity",
+                    to_attr="_identities_main",
```
mjeammet, line 206: why change ?

sampaccoud: the problem with `main_identity` is that you make people believe that it is an identity... and then you have to write `user.main_identity[0]` and it gets weird. In fact this attribute is a subquery so `identities_main` best conveys the fact that there are many identities but that I'm interested in the main one here...

src/backend/core/models.py
sampaccoud: Tough questions. Let''s discuss IRL but some elements of answers:
You can have several webhooks per team if you have several remote applications to sync.
Outline has no SCIM provider yet so the url/secret provisionning can be added there... we can do something simple for a v1...
Before going to production we will need to obfuscate the secret so that users only see it once in the UI (when they set it) and never again...

env.d/development/common.dist
```diff
@@ -33,3 +33,6 @@ LOGOUT_REDIRECT_URL=http://localhost:3000
 
 OIDC_REDIRECT_ALLOWED_HOSTS=["http://localhost:8083", "http://localhost:3000"]
 OIDC_AUTH_REQUEST_EXTRA_PARAMS={"acr_values": "eidas1"}
+
+# Webhooks
+TEAM_WEBHOOKS=[{"url": "http://example.com", "secret": "secret", "verify": true}]
```
sampaccoud, line 38: This lies can now be removed

src/backend/core/models.py
```diff
@@ -421,6 +451,30 @@ class Meta:
...
 
+    def save(self, *args, **kwargs):
+        """
+        Override save function to fire webhooks on any addition or update
+        to a team access.
+        """
+        self.team.webhooks.update(status=WebhookStatusChoices.PENDING)
+        with transaction.atomic():
+            instance = super().save(*args, **kwargs)
+            scim_synchronizer.add_user_to_group(self.team, self.user)
```
mjeammet, line 462: shouldn't we call `add_user_to_group` only when team changes ?

src/backend/core/utils/__init__.py
mjeammet, line 1: Should we empty this file ? This JSonValue method is only used in settings we indent on removing.

mjeammet: I removed it.

src/backend/core/models.py
mjeammet: I'm having a hard time writing the corresponding tests 😨

sampaccoud (approved):
2 nitpicks but otherwise GTM!

mjeammet:
I hesitated to link this PR to issue #106 but I'll open a small dedicated PR later.

Issue #127, opened 2024-03-18
🐛(frontend) router problem build mode


AntoLC:
## Bug Report

Next doesn't manage correctly the dynamic route with SPA in build mode. 
If we access a team page without passing by the internal router, we fall to a 404.

They talk about it here: https://github.com/vercel/next.js/issues/54393

[scrnli_3_18_2024_9-40-02 PM.webm](https://github.com/numerique-gouv/people/assets/25994652/9938f539-13ae-440d-a7bb-1e8cfc810439)

PR #153, opened 2024-03-26
🐛(app-desk) router dynamic

AntoLC:
## Purpose

When we use the static mode, Next.js want that we build our team pages at build time. But we can't do that because users can create a team at runtime.
So we redirect thanks to ngnix when a user try to access a team page (`teams/[id]`) when we see that a team page is not found.

See more here: #127

## Proposal

- [x] redirect `teams` on Desk ngnix conf
- [x] redirect `mail-domains` on Desk ngnix conf

## Demo

[scrnli_6_27_2024_5-17-40 PM.webm](https://github.com/numerique-gouv/people/assets/25994652/b0685acf-e162-4035-96a9-c701615c77f2)

### DIFF
--- a/src/frontend/apps/desk/conf/default.conf
+++ b/src/frontend/apps/desk/conf/default.conf
@@ -12,6 +12,10 @@ server {
     error_page 404 /teams/[id]/;
   }
 
+  location /mail-domains/ {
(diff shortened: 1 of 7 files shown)


daproclaima (approved):
Great workaround !

PR #155, opened 2024-03-26
🚑️(staging) fix 404 errors

lebaudantoine:
Recent changes on the staging cluster created a regression. The ingress `className` needs to be specified.


rouja (approved):
LGTM

PR #159, opened 2024-03-26
🩹(frontend) disable submission button while a request is pending

lebaudantoine:
If any request is taking longer than expected, the user could interact with the frontend, thinking the previous submission was not taken into account, and would re-submit a request.

It happened to me while sending an invitation. Replaying the request can either lead to inconsistencies in db for the user, or to errors in requests' response.

I propose to disable all interactive button while submitting a request. It's good enough for this actual sprint, these types of interactivity issues could be improved later on.

Please find an example where I sleep during a few seconds to simulate what could happen.

https://github.com/numerique-gouv/people/assets/45729124/cac979cd-93b3-408c-8069-59622d2a08e7

### DIFF
--- a/src/frontend/apps/desk/src/features/addMembers/components/ModalAddMembers.tsx
+++ b/src/frontend/apps/desk/src/features/addMembers/components/ModalAddMembers.tsx
@@ -54,8 +54,12 @@ export const ModalAddMembers = ({
   const [selectedMembers, setSelectedMembers] = useState<OptionsSelect>([]);
   const [selectedRole, setSelectedRole] = useState<Role>(Role.MEMBER);
   const { toast } = useToastProvider();
-  const { mutateAsync: createInvitation } = useCreateInvitation();
-  const { mutateAsync: createTeamAccess } = useCreateTeamAccess();
+  const { mutateAsync: createInvitation, isPending: isPendingInvitation } =
+    useCreateInvitation();
+  const { mutateAsync: createTeamAccess, isPending: isPendingTeamAccess } =
+    useCreateTeamAccess();
+
+  const isPending = isPendingTeamAccess || isPendingInvitation;
 
   const switchActions = (selectedMembers: OptionsSelect) =>
     selectedMembers.map(async (selectedMember) => {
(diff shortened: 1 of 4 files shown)


src/frontend/apps/desk/src/features/addMembers/components/ModalAddMembers.tsx
```diff
@@ -54,8 +54,12 @@ export const ModalAddMembers = ({
   const [selectedMembers, setSelectedMembers] = useState<OptionsSelect>([]);
   const [selectedRole, setSelectedRole] = useState<Role>(Role.MEMBER);
   const { toast } = useToastProvider();
-  const { mutateAsync: createInvitation } = useCreateInvitation();
-  const { mutateAsync: createTeamAccess } = useCreateTeamAccess();
+  const { mutateAsync: createInvitation, isPending: isPendingInvitation } =
+    useCreateInvitation();
+  const { mutateAsync: createTeamAccess, isPending: isPendingTeamAccess } =
+    useCreateTeamAccess();
```
AntoLC, line 60: I don't think it is the best way to do, because createInvitation or createTeamAccess can be called multiple times, so if you have multiple members to add, I guess isPending will switch from true to false.

I would set isPending to true here, before the promise stuff:
https://github.com/numerique-gouv/people/blob/b355acb80ae0c69c744799a29cfe6fb3c75ebd9c/src/frontend/apps/desk/src/features/addMembers/components/ModalAddMembers.tsx#L118

And set to false before the onClose:
https://github.com/numerique-gouv/people/blob/b355acb80ae0c69c744799a29cfe6fb3c75ebd9c/src/frontend/apps/desk/src/features/addMembers/components/ModalAddMembers.tsx#L121-L122

AntoLC (commented):
Well seen, you think you could add a Loader as well?
https://github.com/numerique-gouv/people/blob/cc35757c9eea4f6633c4108e8f3d9ada158ad5c0/src/frontend/apps/desk/src/features/teams/components/InputTeamName.tsx#L47-L52

If the tests e2e don't cry I think it is ok to not test that.

AntoLC (approved):
✅ !

PR #162, opened 2024-03-28
📝 add missing info about how to run front in the README

sdemagny:
Just add missing info in the README


AntoLC (approved):
Yes, good thing to do.

The commit message have a nomenclature, to pass the CI do something like that:
📝(frontend) add missing info about how to run front in the README

Issue #107, opened 2024-03-11
Enhance adding members in a team


lebaudantoine:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

The global user search endpoint, utilized for adding new team members, searches the entire User table without excluding current team members. This poses a significant issue, as it allows re-adding members already part of the team, which should not be an option.

**Describe the solution you'd like**

I should have a way to search users excluding all members of a given team.

lebaudantoine:
Related to this feature, we are planning to introduce search within members of a specific team. Soon, users will have the ability to conveniently search for team members using a dedicated search input.

lebaudantoine:
Related to this feature, we should also find a way to avoid inviting a user which is already invited.

PR #163, opened 2024-03-29
✨(backend) enhance search users to add in a team

sdemagny:
Exclude from the result all users already members of the current team

Modify API to only find users not already members of a given team:
https://github.com/numerique-gouv/people/issues/107

### DIFF
--- a/src/backend/core/tests/test_api_users.py
+++ b/src/backend/core/tests/test_api_users.py
@@ -218,6 +218,64 @@ def test_api_users_authenticated_list_by_name_and_email():
     assert user_ids == [str(nicole.user.id), str(frank.user.id), str(david.user.id)]
 
 
+def test_api_users_authenticated_list_exclude_users_already_in_team(
+    django_assert_num_queries,
+):
+    """
+    Authenticated users should be able to search users
+    but the result should exclude all users already in the given team.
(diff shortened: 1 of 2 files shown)


AntoLC (approved):
👍  Seems good to me.

src/backend/core/tests/test_api_users.py
```diff
@@ -218,6 +218,80 @@ def test_api_users_authenticated_list_by_name_and_email():
...
+
+    # We can only find Nicole and Mary because Frank is already a member of the given team
+    nb_queries = 1  # user authenticated
+    nb_queries += 1  # search user query
+    nb_queries += 1  # User
+    nb_queries += 1  # Identity
+    with django_assert_num_queries(nb_queries):
+        response = client.get(
+            f"/api/v1.0/users/?q=ool&team_id={team.id}",
+        )
```
lebaudantoine, line 271: nit-picking: we use a slightly different approach [here](https://github.com/numerique-gouv/people/blob/6d8e05b7464e1ec134078457744c4d6d20e29a2f/src/backend/core/tests/team_accesses/test_api_team_accesses_list.py#L183)

lebaudantoine (approved):

PR #166, opened 2024-03-29
🚨(app-desk) remove warning next.js about css from Head

AntoLC:
## Purpose

Importing the css from Head was causing a flickering effect on the button, because of the time the css
load. Next.js was emitting as well a warning about css being loaded from the Head component.
We moved the css import to the _document.tsx file as recommended by the Next.js documentation.

### Warning
Capture d'écran 2024-03-29 112942

### Flickering
[scrnli_3_29_2024_11-46-32 AM.webm](https://github.com/numerique-gouv/people/assets/25994652/821d44ba-22aa-4479-9b14-595ab6fceee1)

### DIFF
--- a/src/frontend/apps/desk/src/features/header/ApplicationsMenu.tsx
+++ b/src/frontend/apps/desk/src/features/header/ApplicationsMenu.tsx
@@ -1,4 +1,3 @@
-import Head from 'next/head';
 import Script from 'next/script';
 import React from 'react';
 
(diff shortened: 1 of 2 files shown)


manuhabitela:
Cool :) so that's why it flickered… thought Next handled stylesheets that way in any component. Good to know! This was a big work-in-progress to make it work as early as possible, better that way indeed :+1:

manuhabitela (approved):

AntoLC:
> Cool :) so that's why it flickered… thought Next handled stylesheets that way in any component. Good to know! This was a big work-in-progress to make it work as early as possible, better that way indeed 👍

I prefer how the code was before, but yes, it seems to prefer like that.

PR #167, opened 2024-03-29
⚡️(CI) persist the frontend between jobs

AntoLC:
## Purpose

To improve the speed of the CI, we cache the frontend install. 
**It will even be reused between pull request until the yarn.lock has a change.**

We cache as well the desk build app, in another cache, this cache persist only per workflow. 
It will increase the speed if we have e2e flaky tests and that we have to relaunch the e2e job.

⚠️ Firefox e2e tests are often flaky, Firefox is about 3% of the browser used on the web, we remove the e2e support for the moment: https://gs.statcounter.com/browser-market-share

## Archi

image

## Opti

**From 1m54s to 10 seconds.**

image

image

### DIFF
--- a/.github/workflows/people.yml
+++ b/.github/workflows/people.yml
@@ -58,11 +58,9 @@ jobs:
             exit 1
           fi
 
-  test-front:
+  install-front:
     runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: src/frontend/
+
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/people.yml
```diff
@@ -71,40 +69,85 @@ jobs:
...
 
       - name: Test App
-        run: yarn app:test
+        run: cd src/frontend/ && yarn app:test
 
   lint-front:
     runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: src/frontend/
```
lebaudantoine, line 87: A single question, why don't all frontend jobs are running with this default working directory?
```
    defaults:
      run:
        working-directory: src/frontend/
```

lebaudantoine (approved):
Perfecto 👌

.github/workflows/people.yml
AntoLC: It is to avoid confusion, by example here, if you don't specify the path, even with a working directory, it will fail:
https://github.com/numerique-gouv/people/blob/415b86bb97a6c6efd8169819a3f40cbf78570123/.github/workflows/people.yml#L142
So I put everythink with the path, it is a bit redundant, but easy to read and understand.

PR #177, opened 2024-04-05
🧑‍💻(models) improve user str representation

mjeammet:
## Purpose

As suggested in #106 : Improve user model str representation. 

## Implementation

EDIT after Identity model removal  : 
- displays name if user has a name
- displays email if user has no name but an email
- displays sub otherwise

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -230,11 +230,7 @@ class User(AbstractBaseUser, BaseModel, auth_models.PermissionsMixin):
         verbose_name_plural = _("users")
 
     def __str__(self):
-        return (
-            str(self.profile_contact)
-            if self.profile_contact
-            else self.email or str(self.sub)
-        )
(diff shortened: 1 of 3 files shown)


sampaccoud:
@mjeammet  I think technically, on the "sub" field is really needed to identify the user. That was the idea behind the decision to make the other fields like `name` and `email` optional.

mjeammet:
> @mjeammet I think technically, on the "sub" field is really needed to identify the user. That was the idea behind the decision to make the other fields like `name` and `email` optional.

I wonder if the sub isn't necessarily linked to an email. Do we want to allow linking a sub to an adress email different than the one used right before, to log in ? Actually, I guess we do. What do you think ?

mjeammet:
In the end, I moved the email/name/none complexity to Identity's representation. 
In the user model, I return str(main_identity) if there is one, and admin_email / uuid otherwise.

sdemagny (approved):

PR #184, opened 2024-04-09
Logout backend

### LINKED ISSUE #149: Logout session from the OIDC provider
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Please take a look at the mozzila django oidc library, [about the logout](https://mozilla-django-oidc.readthedocs.io/en/stable/installation.html#log-user-out-of-the-openid-connect-provider). Currently, the user logs out from its Django session, but keeps it OIDC provider's one alive.

**Describe the solution you'd like**
Logout from the OIDC provider session following the documentation example.

lebaudantoine:
## Purpose

End OIDC provider session while logging out.

## Proposal

It closes #149

- [x] Improve test Docstring
- [x] Fix tests frontend (units and e2e)
- [x] Fix tests backend

### DIFF
--- a/src/backend/core/tests/authentication/test_views.py
+++ b/src/backend/core/tests/authentication/test_views.py
@@ -1,0 +1,238 @@
+"""Unit tests for the Authentication Views."""
+
+from unittest import mock
+from urllib.parse import parse_qs, urlparse
+
+from django.contrib.auth.models import AnonymousUser
+from django.contrib.sessions.middleware import SessionMiddleware
+from django.core.exceptions import SuspiciousOperation
+from django.test import RequestFactory
+from django.test.utils import override_settings
+from django.urls import reverse
+from django.utils import crypto
+
+import pytest
+from rest_framework.test import APIClient
+
+from core import factories
+from core.authentication.views import OIDCLogoutCallbackView, OIDCLogoutView
+
+pytestmark = pytest.mark.django_db
+
+
+@override_settings(LOGOUT_REDIRECT_URL="/example-logout")
+def test_view_logout_anonymous():
+    """Anonymous users calling the logout url,
+    should be redirected to the specified LOGOUT_REDIRECT_URL."""
+
+    url = reverse("oidc_logout_custom")
+    response = APIClient().get(url)
+
+    assert response.status_code == 302
+    assert response.url == "/example-logout"
+
+
+@mock.patch.object(
+    OIDCLogoutView, "construct_oidc_logout_url", return_value="/example-logout"
+)
+def test_view_logout(mocked_oidc_logout_url):
+    """Authenticated users should be redirected to OIDC provider for logout."""
+
+    identity = factories.IdentityFactory()
+    user = identity.user
+
+    client = APIClient()
+    client.force_login(user)
+
+    url = reverse("oidc_logout_custom")
+    response = client.get(url)
+
+    mocked_oidc_logout_url.assert_called_once()
+
+    assert response.status_code == 302
+    assert response.url == "/example-logout"
+
+
+@override_settings(LOGOUT_REDIRECT_URL="/default-redirect-logout")
+@mock.patch.object(
+    OIDCLogoutView, "construct_oidc_logout_url", return_value="/default-redirect-logout"
+)
+def test_view_logout_no_oidc_provider(mocked_oidc_logout_url):
+    """Authenticated users should be logged out when no OIDC provider is available."""
+
+    identity = factories.IdentityFactory()
+    user = identity.user
+
+    client = APIClient()
+    client.force_login(user)
+
+    url = reverse("oidc_logout_custom")
+
+    with mock.patch("mozilla_django_oidc.views.auth.logout") as mock_logout:
+        response = client.get(url)
+        mocked_oidc_logout_url.assert_called_once()
+        mock_logout.assert_called_once()
+
+    assert response.status_code == 302
+    assert response.url == "/default-redirect-logout"
+
+
+@override_settings(LOGOUT_REDIRECT_URL="/example-logout")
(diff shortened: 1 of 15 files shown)


src/frontend/apps/e2e/__tests__/app-desk/header.spec.ts
```diff
@@ -35,21 +35,17 @@ test.describe('Header', () => {
     await expect(header.getByText('My account')).toBeVisible();
   });
 
-  test('checks logout button', async ({ page }) => {
+  test('checks logout button is visible', async ({ page }) => {
```
lebaudantoine, line 38: I would suggest adding back this test when the e2e will run in a K8s setup using Agent Connect integration environment as OIDC provider. I can open an issue.

AntoLC (commented):
I have a question, could the `logout` route (backend side) kill the session cookie `people_sessionid` ?
Like that the logout button will work as well in development mode.

src/frontend/apps/e2e/__tests__/app-desk/header.spec.ts
sampaccoud: :ok_hand:

src/backend/core/tests/authentication/test_views.py
```diff
@@ -0,0 +1,211 @@
...
+
+from core import factories
+from core.authentication.views import OIDCLogoutCallbackView, OIDCLogoutView
+
+pytestmark = pytest.mark.django_db
+
+
+@override_settings(LOGOUT_REDIRECT_URL="/example-logout")
+def test_view_logout_anonymous():
+    """Anonymous users should be redirected to the specified LOGOUT_REDIRECT_URL."""
```
sampaccoud, line 25:
```suggestion
    """Anonymous users calling the logout url should be redirected to the specified LOGOUT_REDIRECT_URL."""
```

src/backend/core/tests/authentication/test_views.py
```diff
@@ -0,0 +1,211 @@
...
+
+    mocked_oidc_logout_url.assert_called_once()
+
+    assert response.status_code == 302
+    assert response.url == "/example-logout"
+
+
+@override_settings(LOGOUT_REDIRECT_URL="/example-logout")
+def test_view_logout_callback_anonymous():
+    """Anonymous users should be redirected to the specified LOGOUT_REDIRECT_URL."""
```
sampaccoud, line 57: To differentiate from the above test:
```suggestion
    """Anonymous users calling the logout callback url should be redirected to the specified LOGOUT_REDIRECT_URL."""
```

src/backend/core/tests/authentication/test_views.py
```diff
@@ -0,0 +1,211 @@
...
+    assert params["id_token_hint"][0] == "mocked_oidc_id_token"
+    assert params["state"][0] == "mocked_state"
+
+    url = reverse("oidc_logout_callback")
+    assert url in params["post_logout_redirect_uri"][0]
+
+
+@override_settings(LOGOUT_REDIRECT_URL="/")
+def test_view_logout_construct_oidc_logout_url_none_id_token():
+    """If no ID token is available, user should be redirected to the final URL."""
```
sampaccoud, line 133:
```suggestion
    """If no ID token is available in the session, the user should be redirected to the final URL."""
```

src/backend/core/tests/authentication/test_views.py
```diff
@@ -0,0 +1,211 @@
...
+
+    with mock.patch("mozilla_django_oidc.views.auth.logout") as mock_logout:
+
+        def clear_user(request):
+            # Assert state is cleared prior to logout
+            assert request.session["oidc_states"] == {}
+            request.user = AnonymousUser()
+
+        mock_logout.side_effect = clear_user
+        response = callback_view(request)
```
sampaccoud, line 207: I think I would call the callback_view with a newly built request on which I wouldn't set the user and that would be closer to what happens irl where the 2 requests are completely separate?

sampaccoud (approved):

src/frontend/apps/desk/src/core/auth/useAuthStore.tsx
```diff
@@ -1,13 +1,19 @@
...
     new URL('authenticate/', process.env.NEXT_PUBLIC_API_URL).href,
   );
 };
 
+export const lgout = () => {
+  window.location.replace(
+    new URL('logout/', process.env.NEXT_PUBLIC_API_URL).href,
+  );
+};
+
```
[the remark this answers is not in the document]
lebaudantoine: Good catch, feels weird the linter doesn't raise any warning/errors for unused object.

src/backend/core/tests/authentication/test_views.py
lebaudantoine: Let's discuss it when you are back, I am not sure what you mean. I cannot remove the user props, the `WSGIRequest` object requires it. I might be wrongly using it.

lebaudantoine:
I have updated the logout behavior to follow your suggestion @AntoLC! A test in the backend was added, and the e2e test reverted to their original content 🤗

src/frontend/apps/desk/src/core/auth/useAuthStore.tsx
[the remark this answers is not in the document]
AntoLC: The linter doesn't raise anything because you have an export on you function.

src/backend/core/tests/authentication/test_views.py
sampaccoud: ok then you can keep line 294 to set the user to AnonymousUser but start from a new request and not reuse the request that already went through a cycle.

PR #199, opened 2024-04-24
🐛(api) paginate mailbox_manager api results

mjeammet:
## Purpose

Paginate API results of mail-domains and mailboxes endpoints, using Pagination from core.viewsets.

## Proposal

Description...

- [x] user core.viewsets Pagination class
- [x] add tests


jblagadec:
@sdemagny A reformuler + préciser les impacts pour le Front

mjeammet:
Est-ce que c'est toujours un besoin, ça ? @jblagadec

mjeammet:
We'll come back to it when needed.

PR #200, opened 2024-04-25
✨(backend) allow to filter member on team access endpoint

sdemagny:
## Purpose

Filter member by main identity name or email. 

## Proposal

Description...

- [x] filter members using trigram similarity
- [x] add tests

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -347,6 +347,23 @@ class TeamAccessViewSet(
         queryset = queryset.filter(team=self.kwargs["team_id"])
 
         if self.action in {"list", "retrieve"}:
+            if query := self.request.GET.get("q", ""):
+                similarity = Max(
+                    TrigramSimilarity(
+                        Coalesce(Func("user__email", function="unaccent"), Value("")),
+                        Func(Value(query), function="unaccent"),
+                    )
+                    + TrigramSimilarity(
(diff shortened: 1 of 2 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -355,6 +355,26 @@ def get_queryset(self):
...
+                        Func("user__identities__name", function="unaccent"), Value("")
+                    ),
+                    Func(Value(query), function="unaccent"),
+                )
+            )
+            queryset = (
+                queryset.annotate(similarity=similarity)
+                .filter(similarity__gte=SIMILARITY_THRESHOLD)
+                .order_by("-similarity")
+            )
```
mjeammet, line 377: shouldn't it be a little lower, inside the list/retrieve "if" ?

src/backend/core/tests/team_accesses/test_api_team_accesses_list.py
```diff
@@ -124,6 +124,145 @@ def test_api_team_accesses_list_authenticated_related():
...
+        ],
+    )
+    factories.TeamAccessFactory.create_batch(4)
+
+    # Search users in the team Odyssey
+    client = APIClient()
+    client.force_login(user)
+
+    response = client.get(
+        f"/api/v1.0/teams/{team.id!s}/accesses/?q=BRU",
```
mjeammet, line 260: this doesn't work and I'm not sure why 🤔

mjeammet (approved):

PR #209, opened 2024-05-03
♻️(app-desk) adapt La Gaufre

AntoLC:
## Purpose

Adapt La Gaufre with the new configuration.

## Demo

image


manuhabitela:
Hey, I updated the script hosted on the API, it should get rid of the keyboard-related error in the console. Let me know if it's good then!

manuhabitela (approved):

PR #234, opened 2024-06-03
👷(helm) command createsuperuser

AntoLC:
## Purpose

We need a superuser in the Django application, to have access to the admin part.
This PR create a superuser on the pods.

### DIFF
--- a/src/helm/env.d/dev/values.desk.yaml.gotmpl
+++ b/src/helm/env.d/dev/values.desk.yaml.gotmpl
@@ -57,6 +57,13 @@ backend:
     - "people.wsgi:application"
     - "--reload"
 
+  createsuperuser:
+    command:
+      - "/bin/sh"
+      - "-c"
+      - python manage.py createsuperuser --admin_email admin@example.com --password admin
+    restartPolicy: Never
+
 frontend:
   envVars:
     PORT: 8080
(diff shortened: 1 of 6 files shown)


src/helm/env.d/dev/values.desk.yaml.gotmpl
```diff
@@ -57,6 +57,14 @@ backend:
     - "people.wsgi:application"
     - "--reload"
 
+  createsuperuser:
+    command:
+      - "/bin/sh"
+      - "-c"
+      - |
+        python manage.py createsuperuser --admin_email admin@example.com --password admin
```
lebaudantoine, line 65: out of curiosity, why do we need a pipe there? 
Could a code as follows work?
```
-  python manage.py createsuperuser --admin_email admin@example.com --password admin
```

lebaudantoine (approved):
Everything looks good to me, though I'm not an expert with Helm. Could we get @rouja to review this as well?

was It copied from Impress?

I was also wondering if we need to repeat the command declaration in each environment's values.yaml file. Is there a way to avoid duplication across dev, staging, preprod, and prod environments?

It might be a bit early to optimize for this, but I thought I'd bring it up!

AntoLC:
> Everything looks good to me, though I'm not an expert with Helm. Could we get @rouja to review this as well?
> 
> was It copied from Impress?
> 
> I was also wondering if we need to repeat the command declaration in each environment's values.yaml file. Is there a way to avoid duplication across dev, staging, preprod, and prod environments?

Yes it is from Impress, review by @rouja in Impress. 
We need to repeat the cmd because the secret change in every environment.

PR #241, opened 2024-06-04
✨(mailbox_manager) modify API to get maildomain

sdemagny:
Access to maildomain by slug name


src/backend/mailbox_manager/migrations/0005_alter_maildomain_slug.py
mjeammet, line 1: Do you think we could merge the 3 migrations files into one ?

src/backend/mailbox_manager/models.py
```diff
@@ -25,6 +27,11 @@ class Meta:
     def __str__(self):
         return self.name
 
+    def save(self, *args, **kwargs):
+        if not self.slug:
+            self.slug = slugify(self.name)
+        return super().save(*args, **kwargs)
```
mjeammet, line 33: I'm curious about the type of special caracters we would expect in a domain name !

src/backend/mailbox_manager/migrations/0005_alter_maildomain_slug.py
sdemagny: It's important to keep those 3 migrations. When you create a new column with unique value, you have to create first a new empty column then fill the column in new migration and finally turn that column to unique.

mjeammet (approved):

PR #248, opened 2024-06-09
♻️(models) remove multiple identities

sampaccoud:
## Purpose

Multiple identities were complicating this project's code.
We moved the management of multiple identities to our OIDC provider and won't need it anymore here.

## Proposal

Remove the `Identity` model and refactor everything around.


AntoLC (approved):
I find it's a good refacto, simpler now.

src/backend/core/tests/test_models_invitations.py
mjeammet, line 1: Can't comment on line 264 but it fails on my local setting. How come it does not fail here ?

src/backend/core/admin.py
mjeammet, line 1: Access to admin is broken. I'm trying to fix it and commit a fixup before the end of the week.

src/backend/core/models.py
```diff
@@ -205,7 +221,7 @@ class User(AbstractBaseUser, BaseModel, auth_models.PermissionsMixin):
 
     objects = auth_models.UserManager()
 
-    USERNAME_FIELD = "admin_email"
+    USERNAME_FIELD = "sub"
```
mjeammet, line 224: Results in 
image

While it's accurate, I'd like to customize it to "Sub or username" to make it a bit clearer.

src/backend/core/admin.py
mjeammet, line 1: About user creation in the admin : should we deactivate it ? 
If we chose to allow it, we'll have to fill the "sub" field, manually as a field or by generating some random string. This would create users with "wrong" subs, unknown to AgentConnect. 

I say we deactivate it, WDYT @sampaccoud

src/backend/core/tests/test_models_invitations.py
sampaccoud: try running make bootstrap to make sure you are in sync on dependencies?
Can you copy the error message if it persists?` Or we can see today IRL!

src/backend/core/admin.py
sampaccoud: yeah I thought of it and it's actually ok as long as subs are unique.
In our case, you don't risk choosing a sub that will conflict with the IDP... unless you purposely do it but you're an admin so... don't do that?

sampaccoud: I made the modiffication so the "sub" field is now editable on create and readonly after that...

src/backend/core/models.py
sampaccoud: as discussed IRL we keep it like this as it is very "insider"

src/backend/core/tests/test_models_invitations.py
sampaccoud: As discussed IRL, the error was not linked to this PR.

PR #253, opened 2024-06-17
🚧(app-regie) use mail domain slug for mail-domains navigation and requests

daproclaima:
# Stop using mailDomain.id in frontend navigation and requests to the backend. Instead, use mailDomain.url as a slug.

## A meaningful URL, easy to memorize for the user consulting a specific mail domain view. 
With this change, instead of providing a UUID in the frontend application URL, we use the slug of a mail domain.
As a result, a URL that used to have the format `http://localhost:3000/mail-domains/6bbd8081-563a-4c5a-ae02-2785d18f2480` will look like `http://localhost:3000/mail-domains/domainfr`.

## Changes in the code
Each navigation button supposed to redirect to the view of a specific mail domain will embed the slug of the mail domain instead of its id. Also, every http request sent to the django API on the api/mail-domains/ path related to a specific domain and its mailboxes will use the mail domain slug instead of its id (as introduced in commit b4bafb6efbcfae4ed8a116061ee471a27737c41e).

### DIFF
--- a/src/backend/mailbox_manager/api/serializers.py
+++ b/src/backend/mailbox_manager/api/serializers.py
@@ -26,6 +26,12 @@ class MailDomainSerializer(serializers.ModelSerializer):
             "created_at",
             "updated_at",
         ]
+        read_only_fields = [
+            "id",
+            "slug",
+            "created_at",
+            "updated_at",
+        ]
 
 
 class MailDomainAccessSerializer(serializers.ModelSerializer):
--- a/src/backend/mailbox_manager/api/serializers.py
+++ b/src/backend/mailbox_manager/api/serializers.py
@@ -22,6 +22,7 @@ class MailDomainSerializer(serializers.ModelSerializer):
         fields = [
             "id",
             "name",
+            "slug",
             "created_at",
             "updated_at",
         ]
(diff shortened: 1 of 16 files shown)


sampaccoud:
What's the idea behing this change? Can you provide more details in the description of the pull request with our usual format : Purpose / Proposal?

mjeammet:
cannot lint locally, I don't understand why 😢

mjeammet (approved):

src/backend/mailbox_manager/api/serializers.py
```diff
@@ -16,16 +16,29 @@ class Meta:
...
+            "id",
+            "slug",
             "created_at",
             "updated_at",
         ]
 
+    def get_slug(self, instance):
+        """Return slug from the domain's name."""
+        return instance.get_slug()
+
```
AntoLC, line 41: Why do you need to override `slug` ? 
When you save your slug, you already use slugify I can see.
https://github.com/numerique-gouv/people/blob/dc53ecc3bc2ca5d11451ecff8af035fdf3f85230/src/backend/mailbox_manager/models.py#L32-L33

AntoLC (approved):
All seems good ^^

Just a question for @mjeammet.

src/backend/mailbox_manager/api/serializers.py
mjeammet: Actually you're right, I don't think we need that. I removed it.

PR #235, opened 2024-06-03
👷(helm) preprod

AntoLC:
## Purpose

Configure the pre-production environment: https://desk-preprod.beta.numerique.gouv.fr/

Add the deploy workflow, this workflow will deploy the application to the selected tag:
Ex (https://github.com/numerique-gouv/people/releases): 
- Create a release with the version that you want (ex: v0.1.1), set it on the branch/commit that you want to deploy
- It will generate and push the docker images on docker hub - ⚠️ wait that is done (see CI job)
- Remove the `preprod` tag if it is already existing
- Set a new `preprod` tag to the version you just created (v0.1.1) and the correct branch.

Et voila, your preproduction will target the v0.1.1

## Proposal

- [x] 🔧(backend) activate https on oidc redirection
- [x] 👷(helm) preprod configuration
- [x] 👷(CI) add deploy workflow
- [x] 💚(CI) remove trigger workflow on push tags

## Demo

I created a demo branch / PR:
https://github.com/numerique-gouv/people/pull/236

I created the version `v0.1.0-demo`, see https://github.com/numerique-gouv/people/releases/tag/v0.1.0-demo.
I created the tag preprod that target the branch `demo/preprod`, see https://github.com/numerique-gouv/people/releases/tag/preprod

To put to production, we will just have to create our tag `production` that target the branch that we want to release, probably `main`.

PR #258, opened 2024-06-19
🚨(backend) fix warning SSL

AntoLC:
## Purpose

The logs were showing a warning about the SSL:
-  You have not set a value for the SECURE_HSTS_SECONDS setting.
- Your SECURE_SSL_REDIRECT setting is not set to True.

This commit fixes the warning by setting the SECURE_HSTS_SECONDS to 31536000 and SECURE_SSL_REDIRECT to True.

```
security.W004: You have not set a value for the SECURE_HSTS_SECONDS setting. If your entire site is served only over SSL, you may want to consider setting a value and enabling HTTP Strict Transport Security. Be sure to read the documentation first; enabling HSTS carelessly can cause serious, irreversible problems.
security.W008: Your SECURE_SSL_REDIRECT setting is not set to True. Unless your site should be available over both SSL and non-SSL connections, you may want to either set this setting True or configure a load balancer or reverse-proxy server to redirect all connections to HTTPS.
```

## Demo

image

### DIFF
--- a/.github/workflows/docker-hub.yml
+++ b/.github/workflows/docker-hub.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - 'main'
+      - 'fix/warning-backend'
     tags:
       - 'v*'
   pull_request:
--- a/src/backend/people/settings.py
+++ b/src/backend/people/settings.py
@@ -515,6 +515,12 @@ class Production(Base):
     CSRF_TRUSTED_ORIGINS = values.ListValue([])
     SECURE_BROWSER_XSS_FILTER = True
     SECURE_CONTENT_TYPE_NOSNIFF = True
+    SECURE_HSTS_SECONDS = values.IntegerValue(
+        default=31536000, environ_name="SECURE_HSTS_SECONDS"
+    )
+    SECURE_SSL_REDIRECT = values.BooleanValue(
+        default=True, environ_name="SECURE_SSL_REDIRECT"
+    )
 
     # SECURE_PROXY_SSL_HEADER allows to fix the scheme in Django's HttpRequest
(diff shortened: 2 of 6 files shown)


Morendil:
Hi @AntoLC I'm not sure what to do about this PR - what's to be done to get from Draft to Ready to review ? I think I understand the changes to `settings.py` from the first commit but I'm not sure what motivated the changes in the later two commits. Does the `SECURE_SSL_REDIRECT` change have an impact on Helm deployments  ?

mjeammet:
Seems fixed to me. WDYT @AntoLC  ?
image

Morendil:
@mjeammet Ah, I understand why now. The warnings were silenced in [staging](https://github.com/numerique-gouv/people/blob/989239082e99f9e07434710c3f76f4af2444a8ff/src/helm/env.d/staging/values.desk.yaml.gotmpl#L39), [preprod](https://github.com/numerique-gouv/people/blob/989239082e99f9e07434710c3f76f4af2444a8ff/src/helm/env.d/preprod/values.desk.yaml.gotmpl#L39) and [prod](https://github.com/numerique-gouv/people/blob/989239082e99f9e07434710c3f76f4af2444a8ff/src/helm/env.d/production/values.desk.yaml.gotmpl#L39) via PR #235 and #236.

Morendil:
Given #432 I think we can safely close this, however the settings chosen in #432 are not adequate for production uses, where a one- or two-year period is considered best practice. On the other hand, in production these values are normally set by the load balancer or proxies, and this is [apparently](https://domsignal.com/test/v28vg7ht05143uc24z1pi3nlx1gvdinu) the case in our current setup.

PR #269, opened 2024-06-28
🏷️(frontend) adapt ordering name

AntoLC:
## Purpose

A recent change on the backend side ([#2ec29](https://github.com/numerique-gouv/people/commit/2ec292b)) changes the naming ordering params with a prefix "user__". 
We adapt the ordering name in the frontend to match the backend in order to have the sorting working.

## Proposal

- [x] fix sorting
- [x] fix flickering on sorting

## Demo

[scrnli_6_28_2024_10-16-13 AM.webm](https://github.com/numerique-gouv/people/assets/25994652/50d3f633-4a46-43df-86ec-43e72680f3d9)

### DIFF
--- a/src/frontend/apps/desk/src/features/teams/member-management/__tests__/MemberGrid.test.tsx
+++ b/src/frontend/apps/desk/src/features/teams/member-management/__tests__/MemberGrid.test.tsx
@@ -221,10 +221,10 @@ describe('MemberGrid', () => {
   });
 
   it.each([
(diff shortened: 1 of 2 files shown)


daproclaima (approved):
Nice! 
We can plan to have a reflection with @jblagadec about showing the datagrid loader or not when a filtering order request is emitted for the first time. But I think it's not worth it at the moment.

PR #284, opened 2024-07-01
Feature/html lang

daproclaima:
## Purpose

We noticed that the `lang` attribute of the HTML tag does not update when the user changes the language with the language picker, which undermines accessibility.

## Proposal

- [x] update HTML tag when i18next client detects a language change at user input
- [x] include in HTML tag a default value for lang attribute according to browser language (amid supported languages)


AntoLC (approved):
✅ Seems good !

PR #306, opened 2024-07-22
🔨(make) add domains to demo

mjeammet:
## 🎯  Purpose

"Make demo" is very convenient. I use it on a regular basis when resetting my containers. Sole issue, it doesn't create domains and domain accesses. Let's fix that !

## 🔨  Implementation

Add domains creation to "make demo" command. Update related test.

### DIFF
--- a/src/backend/demo/management/commands/create_demo.py
+++ b/src/backend/demo/management/commands/create_demo.py
@@ -10,12 +10,15 @@ from uuid import uuid4
 from django import db
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
+from django.utils.text import slugify
 
 from faker import Faker
 
 from core import models
 
 from demo import defaults
+from mailbox_manager import models as mailbox_models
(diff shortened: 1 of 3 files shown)


src/backend/demo/management/commands/create_demo.py
```diff
@@ -152,6 +155,36 @@ def create_demo(stdout):
...
+                    status=random.choice(MailDomainStatusChoices.choices)[0],
+                )
+            )
+        queue.flush()
+
+    with Timeit(stdout, "Creating accesses to domains"):
+        domains_ids = list(
+            mailbox_models.MailDomain.objects.values_list("id", flat=True)
+        )
+        users_ids = list(models.User.objects.values_list("id", flat=True))
```
sampaccoud, line 177: I would say that users_ids is already set to this as it was already done on line 146 above?

mjeammet: hehe, you're right. I'll remove it right away

sampaccoud (approved):
:tada:

PR #312, opened 2024-08-01
🩺(coverage) add config and make rule to compute coverage

mjeammet:
## Purpose

Compute test coverage to make sure we test every line, especially upon adding a new feature.

## Proposal

- [x] Configure pytest-cov settings in pyproject.toml
- [x] make rule in Makefile

## Improvements
- Should we add a run of coverage in the CI ? We could set a branch rule stating, say, that a branch cannot be merged if it makes coverage go under, say, 90% on main. WDYT ?

### DIFF
--- a/src/backend/pyproject.toml
+++ b/src/backend/pyproject.toml
@@ -137,3 +137,22 @@ python_files = [
     "test_*.py",
     "tests.py",
 ]
+
+[tool.coverage.run]
+branch = true
+omit = [
+    "*/admin.py", 
+    "*/migrations/*", 
+    "*/tests/*",
+    "*/urls.py",
+    "manage.py",
+    "celery_app.py",
+    "wsgi.py",
+    ]
+
+[tool.coverage.report]
+show_missing = true
+skip_empty = true
+
+[tool.coverage.json]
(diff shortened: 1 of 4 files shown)


src/backend/pyproject.toml
```diff
@@ -137,3 +137,19 @@ python_files = [
     "test_*.py",
     "tests.py",
 ]
+
+[tool.coverage.run]
+branch = true
```
mjeammet, line 142: I suggest using branch coverage instead of statement coverage, as it seems more comprehensive https://coverage.readthedocs.io/en/latest/branch.html

sampaccoud (approved):
Nice :star_struck: 
Maybe we will be able to compute coverage in the CI and add a requirement to make sure that each PR does not degrade the coverage score?

mjeammet:
> Nice 🤩 Maybe we will be able to compute coverage in the CI and add a requirement to make sure that each PR does not degrade the coverage score?

Absolutely ! I wanted to do that but maybe in another PR, because I'm not super fluent in github actions and would rather focus on shipping the domains. I'll open an issue.

PR #313, opened 2024-08-02
WIP: manage maildomain access

sdemagny:
## Purpose

- Restrict mailbox creation to domain owner
- get_abilitites from domain

## Proposal

....

### DIFF
--- a/src/backend/mailbox_manager/api/serializers.py
+++ b/src/backend/mailbox_manager/api/serializers.py
@@ -1,8 +1,12 @@
 """Client serializers for People's mailbox manager app."""
 
-from rest_framework import serializers
+from rest_framework import exceptions, serializers
+from timezone_field.rest_framework import TimeZoneSerializerField
 
-from mailbox_manager import models
+from core.api.serializers import DynamicFieldsModelSerializer
+from core.models import User
+
+from mailbox_manager import enums, models
 
 
 class MailboxSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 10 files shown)


src/backend/mailbox_manager/api/serializers.py
```diff
@@ -33,17 +36,133 @@ class Meta:
...
+        fields = [
+            "id",
+            "email",
+            "language",
+            "name",
+            "timezone",
+            "is_device",
+            "is_staff",
+        ]
+        read_only_fields = ["id", "name", "email", "is_device", "is_staff"]
```
mjeammet, line 148:
```suggestion
from core.api.serializer import UserSerializer
```
I don't think you modified it, correct me if I'm wrong !

src/backend/mailbox_manager/api/viewsets.py
```diff
@@ -38,6 +39,15 @@ class MailDomainViewSet(
     lookup_field = "slug"
     queryset = models.MailDomain.objects.all()
 
+    def get_permissions(self):
+        """User only needs to be authenticated to list accesses"""
```
mjeammet, line 43: - unrelated people should not be able to list domain's accesses
- owners, admins and viewers (if we stick to this notion) should list accesses
- about viewers, I think you're right, it might be a good thing for them to list accesses. One use that comes to my mind is knowing which email/person to contact in order to create a new mail on your domain (if your team don't make it clear, at least La Régie wil)

PR #323, opened 2024-08-06
✅(tests) improve tests for mailbox api

mjeammet:
## Purpose

Small PR to slightly improve tests around mailbox_manager API

Regroup mailbox-related tests 
+ add test domain not existing
+ use enabled domain on tests


sdemagny (approved):
Nickel pour moi !

PR #331, opened 2024-08-08
🔥(frontend) remove temporarily team feature

AntoLC:
## Purpose

We want to make a first realease, but the team feature is not ready yet.
So we will hide it for now by hiding the menu and removing the team related tests.

## Proposal

- [x] remove menu
- [x] change root index with `MailDomains` page index
- [x] rename remove temporarily team related test to not be tested

## Demo

image


daproclaima (approved):
🚢 Go!

src/frontend/apps/desk/.env.development
```diff
@@ -1 +1,2 @@
 NEXT_PUBLIC_API_ORIGIN=http://localhost:8071
+NEXT_PUBLIC_FEATURE_TEAM=true
```
sdemagny, line 2: Est-ce qu'on peut déclarer de manière explicite le feature flag aussi pour tous les autres environnements (preprod, prod) ?

daproclaima (approved):
Works ! Thanks for the updates @AntoLC!

Issue #333, opened 2024-08-08
A user can create a mail domain


daproclaima:

jblagadec:
Recette OK

PR #339, opened 2024-08-12
Feature/mail domains status

daproclaima:
## Purpose

Nothing guarantees that once a mail domain creation form is submitted, it can be created automatically. @sdemagny explained that the domain needs to be verified. 
So, the domain will receive a status with the value 'pending,' 'failed,' 'disabled,' or 'active.'

Allowing the assignment of mailboxes to the domain depends on the status: if the status is 'active,' then an administrator or owner can add mailboxes to it.

## Proposal

- show an alert on the top right corner of the page /mail-domains/<slug> according to the status
- disable the mailbox creation button according to the status


daproclaima:
I am not sure about the tests. I have some style fixes to apply and UX writing updates.

daproclaima:
This PR will be closed: its code is merged in PR #333 as it complements it.

resolved #333

PR #345, opened 2024-08-20
🔧Runtime config for the frontend

AntoLC:
## Purpose

We want to be able to pass environment variables at runtime to the frontend. 

To do so, we created a public endpoint on the backend to share  some configurations.
The frontend will get the conf, then adapt itself depend of them.

In this case we wanted to hide the teams feature but only for the production environments (preprod, production).

## Proposal

- [x] ✨(backend) add public endpoint /api/v1.0/config/
- [x] ✨(frontend) add config provider
- [x] ♻️(frontend) replace env NEXT_PUBLIC_FEATURE_TEAM
- [x] 🔧(helm) deactivate teams feature on productions env
- [x] tests

## Demo

[scrnli_8_20_2024_12-35-34 PM.webm](https://github.com/user-attachments/assets/32fc797e-1235-45ea-8e4b-47d4533018f7)


daproclaima (approved):

sdemagny (approved):
Nickel !

PR #365, opened 2024-08-28
📝 (changelog) fix changelog content

sdemagny:
Adapt changelog content to the format choosen (keepachangelog.com)


mjeammet (approved):

mjeammet (changes requested):
We finally decided to not do that

PR #367, opened 2024-08-29
Configure resource server in staging

lebaudantoine:
## Purpose

Configure resource server in staging

## Description

Refer to the merged #308 PR which describes what's a Resource Server.
These client id / secret are generated manually by Agent Connect Ops.
If at any point you need to update them, please contact kevin masseix


rouja (approved):
LGTM

PR #371, opened 2024-08-30
🗃️(mailboxes) migrate local part field modification

mjeammet:
## Purpose

"+" character was removed from valid characters for mailbox creation. 
This commit adds migration file for database to be up-to-date

## Proposal

Description...

- [] item 1...
- [] item 2...


sdemagny (approved):
👍

PR #377, opened 2024-09-04
🐛(dimail) improve handling of dimail errors on failed mailbox creation

mjeammet:
## Purpose

dimail is called twice when creating a mailbox (once for the token,
and once for the post on mailbox endpoint). we want to clarify
the status_codes and messages of each error to inform user and ease debug

## Implementation 

- [x] catch 403 from token
- [x] catch 403 from /mailboxes/new-address/
- [x] catch all other unexpected errors
- [x] 3 related tests


src/backend/mailbox_manager/tests/api/mailboxes/test_api_mailboxes_create.py
```diff
@@ -359,3 +359,125 @@ def test_api_mailboxes__domain_owner_or_admin_successful_creation_and_provisioni
...
+        )
+
+        response = client.post(
+            f"/api/v1.0/mail-domains/{access.domain.slug}/mailboxes/",
+            mailbox_data,
+            format="json",
+        )
+
+        assert response.status_code == status.HTTP_403_FORBIDDEN
+        assert not models.Mailbox.objects.exists()
```
sdemagny, line 396: Est-ce qu'on veut pas aussi checker le body de la response ici ?

PR #387, opened 2024-09-10
🥅(frontend) improve add member form error handling

### LINKED ISSUE #293: Nom de groupe déjà existant : modifier le message d'erreur
Image

Modifications : 

1) Comme la notion de "Slug" ne parle pas aux utilisateurs, il faut remplacer cette phrase pour qu'elle soit explicite : 
- FR : "Un autre groupe utilise déjà ce nom. Veuillez en saisir un autre."
- EN : "This name is already used for another group. Please enter another one." 

2) Comme pour les autres formulaires, il faut que cette erreur soit liée directement au champ auquel elle fait référence.
=> Utiliser le message d'erreur inclus dans le composant initial pour afficher cette erreur.

3) **[MODIF À DISCUTER]** Sur le motif lui-même de l'erreur : comment se fait-il qu'on me dise que ce Slug existe déjà, alors que je ne le vois pas dans l'interface ?
- Est-ce parce qu'il a existé puis a été supprimé ?
 -Est-ce parce qu'un autre utilisateur l'a créé dans un autre groupe, auquel je n'ai pas accès ?
Dans les 2 cas, je pense qu'il faudrait que ce slug/nom soit à nouveau utilisable...

daproclaima:
## Purpose
The error cause displayed when group creation fails because it already exists uses technical wording as it comes directly from the backend. The front end should capture this error and replace it with a message appealing to the end users.

## Proposal
- Capture the error cause 'Team with this Slug already exists.' returned by POST on `api/v1.0.0/teams/` to show 'This name is already used for another group. Please enter another one.' instead.

### Before

https://github.com/user-attachments/assets/4c65384d-d3b3-4421-a149-b5e913220984

### After

https://github.com/user-attachments/assets/972996e0-6193-4dda-87f4-444c38fc37ea

Closes #293.


src/frontend/apps/desk/src/features/mail-domains/api/useAddMailDomain.tsx
```diff
@@ -46,7 +46,9 @@ export const useAddMailDomain = ({
       onSuccess(data);
     },
     onError: (error) => {
-      onError(error);
+      if (typeof onError === 'function') {
```
AntoLC, line 49: Why adding a control ? `onError` is not always a function ?

src/frontend/apps/desk/src/features/teams/team-management/components/InputTeamName.tsx
```diff
@@ -21,14 +23,53 @@ export const InputTeamName = ({
...
+
+    if (handledCauses?.length) {
+      causes = [...handledCauses];
+    }
+    if (unhandledCauses?.length) {
+      causes = [...causes, ...unhandledCauses];
+    }
+
+    return causes;
+  };
```
AntoLC, line 63: If find it is a bit complicated, what about that ?

```suggestion
  const dictCauses: { [key: string]: string } = {
    'Team with this Slug already exists.': t(
      'This name is already used for another group. Please enter another one.',
    ),
  };
  const causes = error?.cause?.map(
    (cause) =>
      dictCauses[cause] ||
      t(
        'Your request cannot be processed because the server is experiencing an error. If the problem ' +
          'persists, please contact our support to resolve the issue: suiteterritoriale@anct.gouv.fr',
      ),
  );
```

AntoLC (approved):
I still find `parseAPIError` very complicated to replace a string by another string. 

I am not sure what is the answer to that: https://github.com/numerique-gouv/people/pull/387#discussion_r1753347447

daproclaima:
@AntoLC, thanks for the remarks. 

I thought of creating this function to reuse it everywhere on APIErrors handling, whether the developer wants to replace an error cause with another message, and/or execute a callback at its detection, and/or show a custom message on 500 error detection.

At the moment, we directly display the 500 error messages on the UI, which may not be the most user-friendly way to explain to the end users that the error is not their fault and that they'd do well reaching out to us so we fix the error.

I used parseAPIError on InputTeamName.tsx to replace a string with another one, as you say, for one of the client errors we already know. I especially wanted to show a custom error message for 500 types of errors.

If I treated only the client error, I would use a more straightforward function. What do you think?

PR #396, opened 2024-09-11
♿️(frontend) update left nav panel

### LINKED ISSUE #302: Bandeau de gauche : supprimer les boutons pour ne conserver que les liens ?

daproclaima:
## Purpose
The HTML semantics needed improvement: the menu items were buttons wrapped in li, which were themselves wrapped in link. This can cause inaccessibility issues to screen reader tools.

## Proposal
Just using a button inside a li.

### Before

https://github.com/user-attachments/assets/62322316-28e2-47c6-8926-26e3dcb67484

### After

https://github.com/user-attachments/assets/68efb55a-390e-43f0-b0e7-411295e4b778

Closes #302

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ and this project adheres to
 - ✨(frontend) allow group members filtering #363
 - ✨(mailbox) send new mailbox confirmation email #397
 
+### Fixed
+
+- ♿️(frontend) fix left nav panel #396
+
 ### Changed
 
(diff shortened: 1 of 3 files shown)


daproclaima:
The tooltip is fixed !

secrets
AntoLC, line 1: I don't think the `secrets` should be versioned, rebase problem ?

AntoLC (approved):

PR #423, opened 2024-09-23
✨(backend) domain accesses update API

sdemagny:
Allow to update (PUT, PATCH) an access.
Role can be change only to a role available
depending to the authenticated user.


mjeammet (approved):
LGTM 👌🏽

daproclaima:
🎉

PR #429, opened 2024-09-27
✨(ci) add security scan

rouja:
Add a security scan for CVE with trivy

### DIFF
--- a/.github/workflows/docker-hub.yml
+++ b/.github/workflows/docker-hub.yml
@@ -16,6 +16,36 @@ env:
   DOCKER_USER: 1001:127
 
 jobs:
+  trivy-scan:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: "people,secrets"
+      -
+        name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+          token: ${{ steps.app-token.outputs.token }}
+      -
+        name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-hub.yml
```diff
@@ -1,4 +1,5 @@
 name: Docker Hub Workflow
+run-name: Docker Hub Workflow
```
Morendil, line 2: Any particular reason to use `run-name` over `name` ? (I don't think this is a blocker for merging the PR but if there is a practical reason it might be useful to document it…)

Morendil (commented):
I notice that the build jobs frequently fail with the following error message (reformatted for clarity):

```
2024-10-30T10:48:12Z	FATAL	Fatal error
  image scan error:
  scan error:
  scan failed:
  failed analysis:
  analyze error:
  pipeline error:
  failed to analyze layer (sha256:e4e85ef47ed5564337c653adcb6f9f3760ead021d1cd0691a29a1ac57fd6f5b0):
  post analysis error: Unable to initialize the Java DB:
  Java DB update failed: DB download error: oci download error:
  failed to fetch the layer:
  GET https://ghcr.io/v2/aquasecurity/trivy-java-db/blobs/sha256:a18f04a942e33d9ffd8cf1ab4d35a09f88936e0729cdd4171dea0aa15a0e457c:
  TOOMANYREQUESTS: retry-after: 80.883µs, allowed: 44000/minute
```

This is troublesome, we've been trying to reduce noise/instability in the build…

This is a known Trivy issue https://github.com/aquasecurity/trivy/discussions/7538

I don't think the issue this can be fixed within this PR, it will require changing the trivy action.

Morendil:
See #500 for a version of this PR without the fixes applied, showing the Trivy report with CVEs.

Morendil:
After discussion with @mjeammet a good solution would be to make this an optional check in CI. Will look into that.

Morendil (approved):
Extracted the scan from build-and-push so an occasional failure for non-security reasons will not block from merging or require re-running jobs.

PR #432, opened 2024-09-30
🔧(backend) fix configuration to avoid different ssl warning

rouja:
Fix following warning messages :
- You have not set a value for the SECURE_HSTS_SECONDS setting.
- Your SECURE_SSL_REDIRECT setting is not set to True.


mjeammet (approved):
back-end tests OK, LGTM

PR #435, opened 2024-10-01
🔧(bakend) use new scaleway email gateway

rouja:
We modify multiples things :
* settings.py in order to manage the new way to send email with the scaleway gateway
* helm template to manage new mandatory secret
* helm configuration for staging/preprod/production

### DIFF
--- a/src/helm/desk/templates/secrets.yaml
+++ b/src/helm/desk/templates/secrets.yaml
@@ -22,3 +22,9 @@ stringData:
 {{- if .Values.mail_provisioning_api_credentials }}
   MAIL_PROVISIONING_API_CREDENTIALS: {{ .Values.mail_provisioning_api_credentials }}
 {{- end }}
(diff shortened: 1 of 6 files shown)


src/helm/desk/templates/secrets.yaml
```diff
@@ -22,3 +22,9 @@ stringData:
 {{- if .Values.mail_provisioning_api_credentials }}
   MAIL_PROVISIONING_API_CREDENTIALS: {{ .Values.mail_provisioning_api_credentials }}
 {{- end }}
+{{- if .Values.djangoEmailHostUser }}
+  DJANGO_EMAIL_HOST_USER: {{ .Values.djangoEmailHostUser }}
```
sampaccoud, line 26: maybe don't put this as a secret so we get the info without having to decode secrets. Since it's not secret...

mjeammet (approved):
LGTM

PR #444, opened 2024-10-03
🐛(dimail) add status code in case of unexpected error

mjeammet:
## Purpose

add status code in case of unexpected error

## Proposal

Description...

- [] item 1...
- [] item 2...


Morendil (approved):
Should make things easier !

PR #445, opened 2024-10-04
🛂(backend) match email if no existing user matches the sub

mjeammet:
## Purpose

Some OIDC identity providers may provide a random value in the "sub" field instead of an identifying ID. In this case, it may be a good idea to fallback to matching the user on its email field.

Adaptation of https://github.com/numerique-gouv/impress/pull/303

Needs fixing when brainjuice will be available

## Proposal

Description...

- [] item 1...
- [] item 2...

### DIFF
--- a/src/backend/core/authentication/backends.py
+++ b/src/backend/core/authentication/backends.py
@@ -1,6 +1,7 @@
 """Authentication Backends for the People core app."""
 
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.core.exceptions import SuspiciousOperation
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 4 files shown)


src/backend/core/authentication/backends.py
```diff
@@ -105,3 +105,32 @@ def create_user(self, claims):
...
+        name_fields = settings.USER_OIDC_FIELDS_TO_NAME
+        full_name = " ".join(
+            user_info[field] for field in name_fields if user_info.get(field)
+        )
+        return full_name or None
+
+    def get_existing_user(self, sub, email):
+        """Fetch existing user by sub or email."""
+        try:
+            return User.objects.get(sub=sub, is_active=True)
```
qbey, line 120: We should not check the `is_active` here, because the result will be a duplicated user with the same `sub`. We should get the user, then later in the calling function, check the `is_active`.

```suggestion
            return User.objects.get(sub=sub)
```

qbey (approved):
This can be merged as it is, I've made a comment I will create an issue for this, the same behavior is already deployed in `impress` and should also be fixed.

PR #454, opened 2024-10-09
✨(dimail) send domain creation request to dimail

mjeammet:
## Purpose

Send domain creation request to dimail when someone creates a domain in people.

## Proposal

Description...

- [x] write function in dimail client
- [x] write an all-OK test
- [x] check actual dimail response upon sending only the "name" of the domain (and check which features must be enabled for valid domain)
...

### DIFF
--- a/src/backend/mailbox_manager/api/serializers.py
+++ b/src/backend/mailbox_manager/api/serializers.py
@@ -77,6 +77,19 @@ class MailDomainSerializer(serializers.ModelSerializer):
             return domain.get_abilities(request.user)
         return {}
 
+    def create(self, validated_data):
+        """
+        Override create function to fire a request to dimail upon domain creation.
+        """
+        # send new domain request to dimail
+        client = DimailAPIClient()
+        client.send_domain_creation_request(
+            validated_data["name"], self.context["request"].user.sub
+        )
+
+        # no exception raised ? Then actually save domain on our database
+        return models.MailDomain.objects.create(**validated_data)
+
 
 class MailDomainAccessSerializer(serializers.ModelSerializer):
     """Serialize mail domain access."""
--- a/src/backend/mailbox_manager/tests/api/mail_domain/test_api_mail_domains_create.py
+++ b/src/backend/mailbox_manager/tests/api/mail_domain/test_api_mail_domains_create.py
@@ -182,10 +182,9 @@ def test_api_mail_domains__no_creation_when_dimail_duplicate(caplog):
             assert response.json() == {"detail": dimail_error["detail"]}
 
     # check logs
-    assert len(caplog.records) == 2  # 1 + new empty info. To be investigated
     record = caplog.records[0]
     assert record.levelname == "ERROR"
     assert (
         record.message
-        == f"[DIMAIL] {dimail_error['status_code']}: {dimail_error['detail']}"
+        == "[DIMAIL] unexpected error : 409 {'detail': 'Domain already exists'}"
     )
--- a/src/backend/mailbox_manager/tests/api/mail_domain/test_api_mail_domains_create.py
+++ b/src/backend/mailbox_manager/tests/api/mail_domain/test_api_mail_domains_create.py
@@ -2,7 +2,13 @@
 Tests for MailDomains API endpoint in People's app mailbox_manager. Focus on "create" action.
 """
 
+import re
+from logging import Logger
+from unittest import mock
+
 import pytest
+import responses
+from requests.exceptions import HTTPError
 from rest_framework import status
 from rest_framework.test import APIClient
 
@@ -56,13 +62,26 @@ def test_api_mail_domains__create_authenticated():
 
     client = APIClient()
     client.force_login(user)
-    response = client.post(
-        "/api/v1.0/mail-domains/",
-        {
-            "name": "mydomain.com",
-        },
-        format="json",
-    )
+
+    domain_name = "test.domain.fr"
+
+    with responses.RequestsMock() as rsps:
+        rsps.add(
+            rsps.POST,
+            re.compile(r".*/domains/"),
+            body=str(
+                {
+                    "name": domain_name,
+                }
+            ),
+            status=status.HTTP_201_CREATED,
+            content_type="application/json",
+        )
+        response = client.post(
+            "/api/v1.0/mail-domains/",
+            {"name": domain_name, "context": "null", "features": ["webmail"]},
+            format="json",
+        )
     assert response.status_code == status.HTTP_201_CREATED
     domain = models.MailDomain.objects.get()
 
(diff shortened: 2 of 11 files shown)


src/backend/mailbox_manager/api/serializers.py
```diff
@@ -77,6 +77,22 @@ def get_abilities(self, domain) -> dict:
...
+        Override create function to fire a request to dimail upon domain creation.
+        """
+        # send new domain request to dimail
+        client = DimailAPIClient()
+        response = client.post_domain_creation_request(
+            validated_data["name"], self.context["request"].user.sub
+        )
+
+        if response.status_code == status.HTTP_201_CREATED:
+            # actually save mailbox on our database
```
sdemagny, line 91: save domain

src/backend/mailbox_manager/utils/dimail.py
```diff
@@ -71,6 +71,48 @@ def get_headers(self, user_sub=None):
 
         return self.pass_dimail_unexpected_response(response)
 
+    def post_domain_creation_request(self, domain_name, request_user):
```
sdemagny, line 74: `send_domain_creation_request` ?

src/backend/mailbox_manager/tests/api/mail_domain/test_api_mail_domains_create.py
```diff
@@ -79,5 +97,51 @@ def test_api_mail_domains__create_authenticated():
...
+    assert rsp.call_count == 1
+
+    # Logger
+    assert mock_info.call_count == 2
+    assert mock_info.call_args_list[0][0] == (
+        "Domain %s successfully created on dimail by user %s",
+        domain_name,
+        user.sub,
+    )
+    # a new empty info has been added. To be investigated
```
sdemagny, line 147: déplacer cette ligne d'info un peu plus haut ligne 140 ?

sdemagny (approved):

src/backend/mailbox_manager/tests/api/mail_domain/test_api_mail_domains_create.py
```diff
@@ -56,13 +61,26 @@ def test_api_mail_domains__create_authenticated():
...
-        "/api/v1.0/mail-domains/",
-        {
-            "name": "mydomain.com",
-        },
-        format="json",
-    )
+
+    domain_name = "test.domain.fr"
+
+    with responses.RequestsMock() as rsps:
```
qbey, line 67: Is there a reason, not to use the responses "classic" way? By classic I mean using the test function decorator `responses.activate` along with 

``` python
    responses.add(
        responses.POST,
        url=re.compile(r".*/domains/"),
        json={
            "name": domain_name,
        },
        status=status.HTTP_201_CREATED,
    )
```

which seems (to me) a bit more readable.

src/backend/mailbox_manager/tests/api/mail_domain/test_api_mail_domains_create.py
```diff
@@ -79,5 +97,93 @@ def test_api_mail_domains__create_authenticated():
 
     # a new domain with status "pending" is created and authenticated user is the owner
     assert domain.status == enums.MailDomainStatusChoices.PENDING
-    assert domain.name == "mydomain.com"
+    assert domain.name == domain_name
     assert domain.accesses.filter(role="owner", user=user).exists()
+
+
+## SYNC TO DIMAIL
+@mock.patch.object(Logger, "info")
```
qbey, line 105: Would you consider using `caplog`? ^^

src/backend/mailbox_manager/utils/dimail.py
```diff
@@ -71,6 +71,48 @@ def get_headers(self, user_sub=None):
...
+        try:
+            response = session.post(
+                f"{self.API_URL}/domains/",
+                json=payload,
+                headers={"Authorization": f"Basic {self.API_CREDENTIALS}"},
+                verify=True,
+                timeout=10,
+            )
+        except requests.exceptions.ConnectionError as error:
+            logger.error(
```
qbey, line 91: As `error` is an exception, why don't you use logger.exception?

mjeammet: Absolutely, I just didn't know there was a dedicated function but that's even better!

About this very line; I might change it for the error to be logged but the response to be passed as is (without exception).

src/backend/mailbox_manager/tests/api/mail_domain/test_api_mail_domains_create.py
mjeammet: We could, although i'd rather discuss/do it in a dedicated PR.

I'm not sold because I'm not sure makes it easier to read. What would be the gain of using caplog, compared to mock ?

src/backend/mailbox_manager/tests/api/mail_domain/test_api_mail_domains_create.py
mjeammet: I haven't had the time to make this PR but I absolutely want to make a fixture to encapsulate these responses ! Do you have a working example of how `responses.activate` functions ?

src/backend/mailbox_manager/tests/api/mail_domain/test_api_mail_domains_create.py
mjeammet: I finally switched to caplog ! I'll have to harmonize uses of caplog on the other tests where we used mock_error

Morendil:
Seems that the only thing we need is to spin up the dimail container, so I propose to pick up this PR. @sdemagny if you want to pair on this ?

Morendil:
Apparently it's not that simple ? I've tried adding the `dimail` container to the runtime dependencies in docker-compose, but we see the domain creation command failing with this error in the `dimail` logs: "You need to init the db by giving me a valid URL".

Further, it seems that the dimail image in the registry hasn't been updated in some time, so in CI we are testing against outdated code. @chrnin Any ideas ?

Morendil:
I think I have a pretty good idea now where this was going wrong. Not one single thing but a bunch of small fixes needed on both Régie and Dimail sides.

- aim to have a "stable" dimail image (as opposed to "latest" which can be too cutting-edge) to code against
- add dimail container to app-dev dependencies in Régie's docker-compose
- use start-dev.sh for the starting command and not entrypoint.sh
- on the Dimail side this start-dev.sh needs to listen on 0.0.0.0 not localhost
- use the more recent schema for the POST /domains request body
- fix the debug logging on Régie side, it's too scarce on details (and breaks owing to trying too hard with UTF-8)
- on the Dimail side the DB was not being fully initialized in FAKE mode

So this needs coordination between an update to this PR and an MR on Dimail.

Stepping back from the nitty-gritty details, a bigger lesson for me is that unless coordination between Régie and Dimail development is more fluid (e.g. by trying cross-team pairing), the lead time for changes of this sort could be intolerably long: weeks to months…

Morendil:
Finally good to go 🎉 after fixing a few remaining tests and changing inter-container communication in CI to use Docker's internal networking between app and Dimail.

PR #461, opened 2024-10-14
✨(frontend) show version number in footer

PanchoutNathan:
## Purpose

In order to see the version number pushed into production

image

### DIFF
--- a/src/frontend/apps/desk/src/core/config/types.ts
+++ b/src/frontend/apps/desk/src/core/config/types.ts
@@ -1,5 +1,6 @@
 export interface Config {
   LANGUAGES: [string, string][];
+  RELEASE: string;
   FEATURES: {
     TEAMS: boolean;
   };
--- a/src/frontend/apps/desk/src/features/footer/Footer.tsx
+++ b/src/frontend/apps/desk/src/features/footer/Footer.tsx
@@ -3,6 +3,7 @@ import { Trans, useTranslation } from 'react-i18next';
 import styled from 'styled-components';
 
 import { Box, LogoGouv, StyledLink, Text } from '@/components';
+import { useConfigStore } from '@/core';
 
 import IconLink from './assets/external-link.svg';
 
@@ -16,6 +17,7 @@ const BlueStripe = styled.div`
 
 export const Footer = () => {
   const { t } = useTranslation();
+  const { config } = useConfigStore();
 
   return (
     <Box $position="relative" as="footer">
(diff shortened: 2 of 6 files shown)


src/frontend/apps/e2e/__tests__/app-desk/footer.spec.ts
```diff
@@ -87,4 +87,31 @@ test.describe('Footer', () => {
...
+              ['en-us', 'English'],
+              ['fr-fr', 'French'],
+            ],
+            FEATURES: { TEAMS: false },
+          },
+        });
+      } else {
+        await route.continue();
+      }
+    });
```
AntoLC, line 108: Actually it could have been interesting to not mock, but to intercept the response of `/api/v1.0/config/` to get the value of `RELEASE`, and then assert on the true backend value, here an interception example:
https://github.com/numerique-gouv/people/blob/579657afa412eb2ce65599de910543c79ac29747/src/frontend/apps/e2e/__tests__/app-desk/member-create.spec.ts#L33-L36

src/frontend/apps/desk/src/features/footer/Footer.tsx
```diff
@@ -138,13 +140,20 @@ export const Footer = () => {
...
         <Text
           as="p"
           $size="m"
           $margin={{ top: 'big' }}
           $variation="600"
           $display="inline"
         >
+          {config?.RELEASE && (
+            <>
+              <Trans>Version</Trans>: {config?.RELEASE} •&nbsp;
```
AntoLC, line 153: I have the feeling it gives more context like that, wdyt ?
```suggestion
              {t(`Version: {{release}}`, { release: config?.RELEASE})} •&nbsp;
```

src/frontend/apps/e2e/__tests__/app-desk/footer.spec.ts
```diff
@@ -87,4 +87,18 @@ test.describe('Footer', () => {
...
+        response.url().includes('/config/') && response.status() === 200,
+    );
+
+    const footer = page.locator('footer').first();
+    await expect(
+      footer.getByText(
+        'Version: NA • Unless otherwise stated, all content on this site is under',
+      ),
+    ).toBeVisible();
+  });
```
AntoLC, line 103: Ah ok, I thought you will get it from a variable from response.
Like that it is fine as well, but you don't need this part so:

```suggestion
  test('check if the app version is visible', async ({ page }) => {
   const footer = page.locator('footer').first();
    await expect(
      footer.getByText(
        'Version: NA • Unless otherwise stated, all content on this site is under',
      ),
    ).toBeVisible();
  });
```

src/frontend/apps/desk/src/core/config/types.ts
```diff
@@ -1,5 +1,6 @@
 export interface Config {
   LANGUAGES: [string, string][];
+  RELEASE?: string;
```
AntoLC, line 3: Can RELEASE be `undefined` actually ?

AntoLC (approved):
✅ Nice !

PR #469, opened 2024-10-17
👔(backend) add `Organization` model

qbey:
## Purpose

We introduce the Organization model has a "hat" for all users and team.

Each User must have a "default" organization.
Each Team must have an organization.

When a User creates a new Team, the team is linked to their default Organization.

For now the Organization should not be visible to end users this is a purely technical aspect as it.

The models are also adding a permission to allow User to edit an Organization, but for now there are no endpoints for that.

## Proposal

Add the Organization model, currently only managed by the backend.

- [x] add `Organization` model
- [x] add FK from `User` to `Organization`
- [x] add FK from `Team` to `Organization`
- [x] add `OrganizationAccess` : not used now
- [x] add "siret" scope (siren is almost never present in IdP)
- [x] Update authentication backend to set organization on user creation
- [x] Update Team creation endpoint to set user organization on the Team
- [ ] **before merge** ask the ProConnect team to allow "siret" scope for "La Régie" SP

Next steps:
- Add an Organization to each User and Team on all environments to mark Organization as mandatory in database.
- Add scope to Organization to list the Service Provider list allowed for a User in an Organization.
- Add endpoints + frontend to manage Organization's scopes


src/backend/people/settings.py
```diff
@@ -335,7 +335,7 @@ class Base(Configuration):
         {}, environ_name="OIDC_AUTH_REQUEST_EXTRA_PARAMS", environ_prefix=None
     )
     OIDC_RP_SCOPES = values.Value(
-        "openid email", environ_name="OIDC_RP_SCOPES", environ_prefix=None
+        "openid email siret", environ_name="OIDC_RP_SCOPES", environ_prefix=None
```
sampaccoud, line 338: should probably be set in the argocd conf only to make the os project more generic?

src/backend/core/models.py
```diff
@@ -158,6 +171,116 @@ def clean(self):
...
+            filters |= models.Q(domains__icontains=domain)
+
+        try:
+            organization = self.get(filters, **kwargs)
+            # If there are several organizations, we must raise an error and fix the data
+        except self.model.DoesNotExist:
+            organization = None
+
+        if organization:
+            return organization, False
```
sampaccoud, line 203:
```suggestion
        try:
            return self.get(filters, **kwargs), False
            # If there are several organizations, we must raise an error and fix the data
        except self.model.DoesNotExist:
            organization = None
```

src/backend/core/models.py
```diff
@@ -158,6 +171,116 @@ def clean(self):
...
+    Each Team have an Organization, which is the Organization from the User who created
+    the Team.
+
+    Organization is managed automatically, the User should never choose their Organization.
+    When creating a User, you must use the `get_or_create` method from the
+    OrganizationManager to find the proper Organization.
+    """
+
+    name = models.CharField(_("name"), max_length=100)
+    sirets = ArrayField(
```
sampaccoud, line 246: Can you say again in the docstring why we can have several sirets or domains for one organization?

src/backend/core/authentication/backends.py
```diff
@@ -101,13 +139,40 @@ def create_user(self, claims):
             raise SuspiciousOperation(
                 _("Claims contained no recognizable user identification")
             )
+        email = claims.get("email")
+        name = claims.get("name")
 
-        return self.UserModel.objects.create(
+        # Extract or create the organization from the data
+        domain = get_domain_from_email(email)
```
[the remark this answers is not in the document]
qbey: It was not DRY on purpose, because I want the code for existing user to be deleted once this is deployed and all existing user have an organization. Not be DRY will allow easy cleanup (and after the cleanup only this part will remain).

src/backend/core/authentication/backends.py
```diff
@@ -90,7 +106,29 @@ def get_or_create_user(self, access_token, id_token, payload):
...
+            domain = get_domain_from_email(email)
+            try:
+                organization, organization_created = Organization.objects.get_or_create(
+                    siret=claims.get("siret"),
+                    domain=domain,
+                )
+                if organization_created:
+                    logger.info("Organization %s created", organization)
+                    # For this case, we don't create an OrganizationAccess we will
+                    # manage this later.
```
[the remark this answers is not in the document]
qbey: It's not handled on purpose, because I don't the first user who login after the release to be the admin of their organization. So we will keep organization without admin and we will have to manually clean things up (while there is not that much organization in the database). I will improve the comment.

src/backend/core/authentication/backends.py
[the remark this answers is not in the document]
sampaccoud: Maybe create an issue for this?

src/backend/core/authentication/backends.py
```diff
@@ -90,7 +111,38 @@ def get_or_create_user(self, access_token, id_token, payload):
                 raise SuspiciousOperation(_("User account is disabled"))
             self.update_user_if_needed(user, claims)
         elif self.get_settings("OIDC_CREATE_USER", True):
-            user = User.objects.create(sub=sub, password="!", **claims)  # noqa: S106
+            user = self.create_user(claims)
+
+        # Data cleaning, to be removed when user organization is null=False
```
sampaccoud, line 116: :medal_military:

src/backend/core/authentication/backends.py
```diff
@@ -90,7 +111,38 @@ def get_or_create_user(self, access_token, id_token, payload):
...
+        # or all users have an organization.
+        if not user.organization_id:
+            organization_registration_id = claims.get(
+                settings.OIDC_ORGANIZATION_REGISTRATION_ID_FIELD
+            )
+            domain = get_domain_from_email(email)
+            try:
+                organization, organization_created = Organization.objects.get_or_create(
+                    registration_id=organization_registration_id,
+                    domain=domain,
```
sampaccoud, line 126: Isn't it good enough if one of the fields matches? I'm thinking for example of a case where the organization ID was not transmitted by the OIDC provider and gets added suddenly.
What do we want to do in case where an organization exists with this domain but it was associated with another ID.

{EDIT] I realized further down the review that you handled it by overidding `get_or_create`. I find it confusing to override a well known method. Could it be a different name or is there a reason why you think it's better to override?  get_or_smart_create? ^^

src/backend/core/factories.py
```diff
@@ -119,13 +119,37 @@ class Meta:
...
+class OrganizationFactory(factory.django.DjangoModelFactory):
+    """Factory to create organizations for testing purposes."""
+
+    name = factory.Faker("company")
+
+    class Meta:
+        model = models.Organization
+
+    class Params:  # pylint: disable=missing-class-docstring
+        with_siret = factory.Trait(
```
sampaccoud, line 131: with_registration_id?

src/backend/core/models.py
```diff
@@ -158,6 +174,133 @@ def clean(self):
...
+
+        # Manage the case where the organization does not exist: we create one
+        if registration_id:
+            return self.create(
+                name=registration_id, registration_id_list=[registration_id], **kwargs
+            ), True
+
+        if domain:
+            return self.create(name=domain, domain_list=[domain], **kwargs), True
+
```
sampaccoud, line 216: from what I understand here, if the OIDC provider gives you both registration_id and domain, you don't store the domain. I think there is a smart reason why you don't save both (something like you are trying to keep minimum identificating information and not maximum information). Maybe you think domain is more fragile and should be given to another organization in the future... ? I'm sure it happens. So you are probably right but can you give us a braindump in comments here?

src/backend/core/tests/authentication/test_backends.py
```diff
@@ -158,25 +161,22 @@ def get_userinfo_mocked(*args):
 
 def test_authentication_getter_new_user_no_email(monkeypatch):
     """
-    If no user matches the user's info sub, a user should be created.
-    User's info doesn't contain an email/name, created user's email/name should be empty.
+    If no user matches the user's info sub, a user should not be created without email nor SIRET.
```
sampaccoud, line 164:
```suggestion
    If no user matches the user's info sub, a user should not be created without email nor organization registration ID.
```

src/backend/core/tests/authentication/test_backends.py
```diff
@@ -158,25 +161,22 @@ def get_userinfo_mocked(*args):
...
     """
     klass = OIDCAuthenticationBackend()
 
     def get_userinfo_mocked(*args):
-        return {"sub": "123"}
+        return {"sub": "123"}  # No email, no SIRET
```
sampaccoud, line 169:
```suggestion
        return {"sub": "123"}  # No email, no organization registration ID
```

src/backend/core/models.py
qbey: I added this note in the docstring:

```
        Dev note: When a registration_id is provided by the Identity Provider, we don't want
        to use the domain to create the organization, because it is less reliable: for example,
        a professional user, may have a personal email address, and the domain would be gmail.com
        which is not a good identifier for an organization. The domain email is just a fallback
        when the registration_id is not provided by the Identity Provider. We can use the domain
        to create the organization manually when we are sure about the "safety" of it.
``` 

I hope this makes it clearer :)

src/backend/core/authentication/backends.py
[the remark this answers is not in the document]
qbey: Issue added https://github.com/numerique-gouv/people/issues/504

src/backend/core/models.py
sampaccoud: This comment will be precious one day! :pray:

sampaccoud (approved):
Let's get organized :tada: !

PR #474, opened 2024-10-21
✅(tests) test dimail setup command

mjeammet:
## Purpose

Test (and slightly improve) dimail setup command, which populate db on local dimail container

## Proposal

Description...

- [] item 1...
- [] item 2...

### DIFF
--- a/src/backend/mailbox_manager/tests/test_command_dimail_setup_db.py
+++ b/src/backend/mailbox_manager/tests/test_command_dimail_setup_db.py
@@ -1,0 +1,69 @@
+"""Test the `setup_dimail_db` management command"""
+
+from django.conf import settings
+from django.contrib.auth import get_user_model
+from django.core.management import call_command
+
+import pytest
+import requests
+
+from mailbox_manager.management.commands.setup_dimail_db import DIMAIL_URL, admin
+
+pytestmark = pytest.mark.django_db
+
+admin_auth = (admin["username"], admin["password"])
+User = get_user_model()
+
+
+@pytest.mark.skipif(
+    settings.DEBUG is not True,
+    reason="Run only in local (dimail container not running in other envs)",
+)
+def test_commands_setup_dimail_db():
+    """The create_demo management command should create objects as expected."""
+    call_command("setup_dimail_db")
+
+    # check created users
+    response = requests.get(url=f"{DIMAIL_URL}/users/", auth=admin_auth, timeout=10)
+    users = response.json()
+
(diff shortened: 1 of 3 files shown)


src/backend/mailbox_manager/tests/test_command_dimail_setup_db.py
```diff
@@ -0,0 +1,69 @@
...
+
+    # check created users
+    response = requests.get(url=f"{DIMAIL_URL}/users/", auth=admin_auth, timeout=10)
+    users = response.json()
+
+    # if John Doe exists, we created a dimail user for them
+    local_user = User.objects.filter(name="John Doe").exists()
+
+    assert len(users) == 3 if local_user else 2
+    # remove uuid because we cannot devine them
```
Morendil, line 34: devine -> guess ;)

Morendil (approved):

src/backend/mailbox_manager/tests/test_command_dimail_setup_db.py
```diff
@@ -0,0 +1,69 @@
...
+    call_command("setup_dimail_db")
+
+    # check created users
+    response = requests.get(url=f"{DIMAIL_URL}/users/", auth=admin_auth, timeout=10)
+    users = response.json()
+
+    # if John Doe exists, we created a dimail user for them
+    local_user = User.objects.filter(name="John Doe").exists()
+
+    assert len(users) == 3 if local_user else 2
```
Morendil, line 33: This `if` is a bit confusing. I'm not 100% sure I understand this test, but if I do understand, it should run only in local environment and only after `setup_dimail_db` has been called ? Perhaps then it would be better to assert a length of 3 without the if. Then the test will fail (as it should) if the user has forgotten to run `setup_dimail_db`.

Morendil: I understand a bit better now - the difference between 3 or 2 users corresponds to whether there has been a connection locally with the login "people" - which is then automatically created with the name "John Doe". So we need the test to be robust, as we might be running it without having connected first.

Morendil: If that's OK I'll take this up and add a commit to make the test more deterministic.

mjeammet: As discussed, @Morendil forces the creation of the local_user before calling the command, thus enforcing tht all 3 users are created.

PR #475, opened 2024-10-21
🐛(script) improve and fix release script

sdemagny:
Use regex to replace version and add missing frontend update version.


mjeammet (approved):
magnifique

PR #480, opened 2024-10-23
🗃️(teams) remove `slug` field

qbey:
## Purpose

After some reflexion, the use of a slug field raises to many problems without being really needed.

One problem is the slug is made from the group name, but we don't have unicity on this, so a user might be blocked without any clue.

We also want to allow group names to be reused (which is already allowed except for the automatic slug).

The unique ID that will be shared with Service Providers will be the PK/UUID.

FIXES https://github.com/numerique-gouv/people/issues/293

## Proposal

Remove the `slug` field of the `Team` model.

- [x] Update `Team` model
- [x] Update related tests
- [x] Add database migration


CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to
 - 🐛(sentry) fix duplicated sentry errors #479
 - 🐛(script) improve and fix release script
 
+### Changed
+
+- 🗃️(teams) remove `slug` field
```
sampaccoud, line 18: There is is "Removed" section https://keepachangelog.com/en/1.0.0/

src/backend/core/migrations/0002_remove_team_slug.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 5.1.1 on 2024-10-23 15:19
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
sampaccoud, line 6: The migration should be delayed to a further release to avoid errors in production during the deployment.
We didn't implement syzygy yet!

sampaccoud (approved):

PR #488, opened 2024-10-24
🐛(sentry) restore default integrations

mjeammet:
## Purpose

a previous commit turned sentry's default_integrations to false to remove duplicates on monitored issues. Thing is, duplicates were appearing because we actually logged + raised the issue where only one of these methods is necessary. Better restore Sentry's default parameters and try to clean our code of duplicate logging

## Proposal

Description...

- [] item 1...
- [] item 2...


qbey:
@mjeammet I've finally understood what the problem is (thanks to @NathanVss 's help). 

We use the `DockerflowMiddleware` whose job is precisely to add a log for each request. So we there is an exception, we catch it and the middleware also adds a log : twice in Sentry.

I think the proper solution is to ignore the `request.summary` logger with:

``` python
from sentry_sdk.integrations.logging import ignore_logger

#...

sentry_sdk.init(
    # ...
)

# Ignore the logs added by the DockerflowMiddleware
ignore_logger("request.summary")
```

With that, we won't have duplicated logs anymore and we will be able to keep the default integrations :)

sdemagny (approved):

PR #491, opened 2024-10-25
Add linter rule for missing docstrings

### LINKED ISSUE #104: Faulty linter
## Bug Report

**Problematic behavior**
Functions or class without docstring pass ruff linter.

Morendil:
## Purpose

Fixes #104

## Proposal

Add D1 (pydocstyle rule set related to missing docstrings) only for the time being. This requires fixing 73 missing docstrings. The full set or pydocstyle rules requires 413 fixes, comprising mostly whitespace and punctuation issues.

We exclude D106 (missing docstring in nested class) owing to heavy use of Meta nested classes in Django. See https://github.com/PyCQA/pydocstyle/pull/261

We exclude D105 (missing docstring in magic method) as it picks up a number of `__str__` methods, which are not high-value docstring targets, the docstring tends to be a restatement of what the single line of code does.

This brings down the total to about 30 missing docstrings, which is reasonable for a single PR.


src/backend/pyproject.toml
```diff
@@ -116,8 +116,12 @@ select = [
     "S", # flake8-bandit
     "SLF", # flake8-self
     "T20", # flake8-print
+    "D1", # pydocstyle
 ]
-ignore= ["DJ001", "PLR2004"]
+ignore= ["DJ001", "PLR2004", "D105", "D106"]
+
+[tool.ruff.lint.pydocstyle]
+convention = "google"
```
mjeammet, line 124:
```suggestion
```

mjeammet (approved):

PR #31, opened 2024-01-29
✨Search group menu

AntoLC:
## Purpose

Integrate the design and partial interactions of the search group menu.

## Proposal

- [x] Integrate the design
- [x] Display teams depend the cases (no teams / teams empty / teams with users)
- [x] Add actions interactions 
- [x] Add Infinite scroll bar
- [x] more tests

## Demo

[scrnli_2_2_2024_4-01-37 PM.webm](https://github.com/numerique-gouv/people/assets/25994652/0c0a6242-8137-431a-abf7-0c59c4067550)

PR #492, opened 2024-10-26
Improve E2E tests: fix stability issues, tighten controls

### LINKED ISSUE #409: Flaky e2e tests
## Bug Report

**Problematic behavior**
Les tests e2e sont un peu flaky. Comme ça n'arrive pas souvent, il suffit de relancer le job mais comme ce job est très long, ça fait quand même 2x10 minutes d'attente.

**Steps to Reproduce**
1. push code to any pull request
2. wait
3. And then the bug happens! (but not everytime)

example of failing test : https://github.com/numerique-gouv/people/actions/runs/10959119350/job/30430871198

Morendil:
## Purpose

At present E2E tests are more of a burden on the team than they are helping with quality, due to instability which I believe was mainly caused by parallel workers, but these problems were hidden by the two-retries policy.

This PR fixes some of the stability issues but also argues for following the testing philosophy **"fail fast"**.

Rather than "retrying flaky tests" I propose seeing any test failure, even intermittent, as a sign of a potential problem. The right thing to do with a potential problem is not, as it were, to sweep it under the rug, but to fix it for the future.

If a test is truly flaky then it needs to be fixed or removed - E2E tests need to pull their weight, given their cost in terms of longer feedback loops.

## Proposal

This PR
- rolls back the "multiple workers in CI " change made in #31 (which is redundant with sharding anyway)
- disables retries - if tests fail, we want to diagnose and fix the failure
- saves Playwright traces (which were super helpful in diagnosing these problems)
- saves server logs alongside traces for ease of debugging
- adds a QR fix to the occasional failure in "Add dummy data", caused by FactoryGirl sometimes creating duplicate domains

Fixes #409 (hopefully)

### DIFF
--- a/.github/workflows/people.yml
+++ b/.github/workflows/people.yml
@@ -203,6 +203,10 @@ jobs:
       - name: Run e2e tests
         run: cd src/frontend/ && yarn e2e:test --shard=${{ matrix.shardIndex }}/${{ matrix.shardTotal }}
 
+      - name: Save logs
+        if: always()
+        run: docker compose logs > src/frontend/apps/e2e/report/logs.txt
+
       - uses: actions/upload-artifact@v4
         if: always()
         with:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to
 
 ### Fixed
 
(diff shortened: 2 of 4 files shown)


mjeammet (approved):
LGTM !

PR #494, opened 2024-10-28
Improve front-end caching

Morendil:
## Purpose

Caching was ineffective on E2E setup, this simplifies caching logic and will hopefully speed up the E2E testing part of the build.

## Proposal

Take advantage of the built-in caching directive in newer versions of the Node setup action.

### DIFF
--- a/.github/workflows/people.yml
+++ b/.github/workflows/people.yml
@@ -67,24 +67,8 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: '18.x'
-
-      - name: Restore the frontend cache
-        uses: actions/cache@v4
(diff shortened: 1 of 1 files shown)


Morendil:
Closing this version because looking at the time where time is actually spent show that I was not going in the right direction. What's taking a long time is not the `yarn install` part but the `playwright install` part. These are system dependencies not Node dependencies… Also the caching system in `actions/setup-node` is substantially (and confusingly) different from `actions/cache`.

Morendil:
In addition, [this bit of bad news](https://playwright.dev/docs/ci#caching-browsers) is kind of putting me off the idea of caching this particular step at all…

PR #496, opened 2024-10-29
🔥(teams) remove pagination of teams listing

sdemagny:
For frontend pagination is useless for teams,
so we remove it.
fixes [#497](https://github.com/numerique-gouv/people/issues/497)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,14 @@ and this project adheres to
 
 ## [Unreleased]
 
-### Added
-
-- ✨(teams) register contacts on admin views 
-
 ### Changed
 
+- 🔥(teams) remove pagination of teams listing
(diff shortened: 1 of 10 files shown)


Morendil (changes requested):
Unfortunately this is incomplete - removing pagination leads to removing the indirection via `results`, but the frontend is not aware of this change, and so just about every page breaks with an exception when trying to access `page.results`.

(Even more unfortunately Playwright - or our testing approach - is not smart enough to recognize an exception is not the right behaviour, and instead Playwright spends the full timeout period waiting for something that will never appear. This is why E2E tests are timing out on this PR. The only good news is that when running the tests locally, this exception is visible in the Playwright trace viewer.)

Morendil (approved):

PR #503, opened 2024-11-04
⬆️(dependencies) remove unneeded dependencies

Morendil:
## Purpose

To reduce attack surface and dependency updates, remove listed dependencies that are not actually in use.

## Proposal

Remove url-normalize which is not referenced anywhere

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,9 @@ and this project adheres to
 
 ## [Unreleased]
 
-### Changed
+### Removed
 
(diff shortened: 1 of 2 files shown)


qbey:
boto3 is used by `django-storages` with S3
Brotli is used by `whitenoise`
`url-normalize` seems indeed not used anywhere :)

Morendil:
To recap our quick in-person chat: for django-storages and whitenoise respectively, boto3 and Brotli are transitive dependencies but marked as optional, so we're leaving them in to ensure they're both picked up.

PR #509, opened 2024-11-06
✨(teams) enable the feature for specific user

qbey:
## Purpose

We want user who are admin/owner of a group to be able to access the management menu.

## Proposal

To enable the Team management feature only to few users, we propose to add a user "abilities" to retrieve whether a user can see Team / Contact / Mailboxes or not, and also allow creation or not.

- [x] Add `User` abilities
- [x] Update the `/me` endpoint to return these abilities
- [x] Update the frontend to take theses new information into account


src/backend/core/api/serializers.py
```diff
@@ -71,6 +71,39 @@ class Meta:
...
+    class Meta:
+        model = models.User
+        fields = [
+            "id",
+            "email",
+            "language",
+            "name",
+            "timezone",
+            "is_device",
+            "is_staff",
```
sampaccoud, line 92: maybe sort alphabetically in this sublist!

src/backend/core/api/viewsets.py
```diff
@@ -188,9 +188,16 @@ class UserViewSet(
     permission_classes = [permissions.IsSelf]
     queryset = models.User.objects.all().order_by("-created_at")
     serializer_class = serializers.UserSerializer
+    me_serializer_class = serializers.UserMeSerializer
```
sampaccoud, line 191: Reminds me that we have this SerializerPerAction mixin: https://github.com/numerique-gouv/people/pull/509/files#diff-cc27a5de6f1d8131250e4be30d0fd4c2f1a816bdc35951a9c86e6f13cd795019R82

src/backend/core/models.py
```diff
@@ -441,6 +441,62 @@ def email_user(self, subject, message, from_email=None, **kwargs):
             raise ValueError("You must first set an email for the user.")
         mail.send_mail(subject, message, from_email, [self.email], **kwargs)
 
+    def get_abilities(self):
+        """
+        Return the user permissions at the application level (ie feature availability).
+
+        Note: this is for display purposes only for now.
```
sampaccoud, line 448: one of the main interests of this pattern was unicity of the information used for display and for check in permissions. Do you think it makes sense here?

src/backend/core/models.py
```diff
@@ -441,6 +441,62 @@ def email_user(self, subject, message, from_email=None, **kwargs):
...
+                ),
+            },
+            "teams": {
+                "canView": teams_can_view,
+                "canCreate": teams_can_view and settings.FEATURES_TEAMS_CREATE,
+            },
+            "mailboxes": {
+                "canView": mailboxes_can_view,
+                "canCreate": mailboxes_can_view and settings.FEATURES_MAILBOXES_CREATE,
+            },
```
sampaccoud, line 497: We should converge to a common format for abilities between our projects... I stick to snake_case and flat format in our other code bases... to be discussed.

src/backend/people/settings.py
```diff
@@ -451,6 +451,19 @@ class Base(Configuration):
...
+    )
+    FEATURES_CONTACTS_CREATE = values.BooleanValue(
+        default=True, environ_name="FEATURES_CONTACTS_CREATE", environ_prefix=None
+    )
+    FEATURES_TEAMS_CREATE = values.BooleanValue(
+        default=True, environ_name="FEATURES_TEAMS_CREATE", environ_prefix=None
+    )
+    FEATURES_MAILBOXES_CREATE = values.BooleanValue(
+        default=True, environ_name="FEATURES_MAILBOXES_CREATE", environ_prefix=None
+    )
```
sampaccoud, line 466: I think we need a better name. I would have prefixed them with USER_ 
USER_MAILBOXES_CREATE ?

sampaccoud (approved):

src/backend/people/settings.py
sampaccoud: And also what about the setting FEATURES below :thinking:

src/backend/demo/management/commands/create_demo.py
```diff
@@ -224,6 +223,75 @@ def create_demo(stdout):
...
+                    )
+                )
+                queue.push(
+                    mailbox_models.MailDomainAccess(
+                        domain_id=domains_ids[0],
+                        user_id=team_mail_user.pk,
+                        role=domain_role,
+                    )
+                )
+
```
sampaccoud, line 294: that's fine but note that the queue pattern was for doing bulk create with tens of millions of users to test scaling. It would be ok to also create these few users with standard calls to factories.

sampaccoud (approved):
:tada: Let's :ship: it!

src/backend/demo/management/commands/create_demo.py
qbey: I keep it like this for now, I was wondering whether I should stick to the existing code or not.

PR #511, opened 2024-11-08
Fix version data generation and display

### LINKED ISSUE #510: Fix version display
## Bug Report

**Problematic behavior**
Version display is constant "NA"

**Expected behavior/code**
Version display should track actual version

Morendil:
## Purpose

Display version information in the app's footer to help acceptance testing and debugging.

Fixes #510 

## Proposal

Improve release script to generate the version.json file that helps the frontend show a version number via the config endpoint, also embed the commit hash directly in frontend to serve as a sanity check in case the frontend fails to build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to
 
 ### Fixed
 
+- 🐛(debug) fix version data generation and display
 - 🐛(mail) fix display button on outlook
 - 💚(ci) improve E2E tests #492
 - 🔧(sentry) restore default integrations
--- a/scripts/release.py
+++ b/scripts/release.py
@@ -5,7 +5,9 @@ import os
 import re
 import sys
 
-from utils import run_command
+import json
+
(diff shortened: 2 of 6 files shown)


qbey:
@Morendil isn't there a way to fetch the version from the package.json file? (cc @PanchoutNathan maybe?)

Morendil:
@qbey There are a [variety of ways](https://stackoverflow.com/questions/9153571/is-there-a-way-to-get-the-version-from-the-package-json-file-in-node-js-code), but it's hard to find one that's entirely satisfactory.

I'd like to take advantage of this PR to also get the commit, because then we can easily tell exactly what we are running in all environments from staging to production, and I'd like to be sure both the frontend and backend are reporting where they're at, given our recent issues with updating the frontend image.

mjeammet (approved):
👏🏽

Morendil:
Closing in favor of an alternative approach, i.e. actually implementing what's documented in `release.py` : generating the version information during Docker image build.

PR #527, opened 2024-11-15
✨(domains) allow creation of "pending" mailboxes

mjeammet:
## Purpose

Previously, mailbox creation was restricted to "active" domains. We now allow users to create mailboxes on pending/failed/disabled domains, and mailboxes thus created have the "pending" mailboxes status. They are sent to dimail upon domain status switching to "active".

## Proposal

Description...

- [x] remove restrictions preventing from creating mailboxes on pending/failed/disabled domains
- [x] document business logic change in docs
- [x] fix tests


mjeammet:
Should we allow mailbox creation for all three types ? Pending / Failed / Disabled domains ?

@sdemagny @Morendil

Morendil:
@mjeammet I'd say yes, always create.

My current working metaphor (open to being replaced by a better one) is "moving into a new home". There is no heating/electricity/water yet (the domain is not connected) but as long as you are in the building, it is comforting and keeps the project moving if you are able to set up the furniture the way you want (creating groups and mailboxes).

I'm not sure exactly what domain state transitions we are effecting automatically, but assuming I understand the intentions:

Pending has the meaning of "there is no electricity yet". Cool, while I wait for that to happen I can keep setting up.

Failed has the meaning of "we tried setting you up with the electricity provider but something went wrong". Well, I'll have to look into what went wrong, but perhaps I don't know what to do and have to leave a message. Meanwhile, I can hope things will get fixed, and keep setting up.

Perhaps Disabled is different.

Disabled maybe has the meaning of "you said you didn't want to live here any more so we're cutting off electricity", and in that case there's no reason to keep rearranging furniture. But perhaps it has another meaning like "this domain was an alternate address for the same building" - if we support domain aliasing. We don't yet know how we're going to handle domain aliases.

But my heuristic in these cases would be to implemented the simplest thing, and the simplest thing is to not pay attention to domain status at all.

sdemagny (commented):
Je rajoute le cas particulier de l'état disable dans un commit que tu squasheras si ça te va ;)

sdemagny:
Dans les deux commits à squasher avec le reste si ça te va Marie :
- [x] gestion de l'interdiction de créer une boite mail lorsque le domaine a été désactivée
- [x] ajout des trads manquantes (il en manque plein mais j'ai juste traduit celles qu'on utilise dans notre PR)
- [x] ajout du status au serializer de la mailbox (le front pourra l'afficher dès maintenant comme ça)

sdemagny (approved):

PR #534, opened 2024-11-19
🐛(mailboxes) fix logic when attempting to duplicate mailboxes

mjeammet:
## Purpose

Logic was broken when attempting to duplicate mailboxes, where infos were sent to dimail before checking for uniqueness.

## Proposal

I'd rather add this logic to a `validate` function in serializer, but that would require some heavier refacto (addition of domain in serializer or at least in context)

### DIFF
--- a/src/backend/mailbox_manager/tests/api/mailboxes/test_api_mailboxes_create.py
+++ b/src/backend/mailbox_manager/tests/api/mailboxes/test_api_mailboxes_create.py
@@ -318,6 +318,57 @@ def test_api_mailboxes__no_dimail_call_if_mailbox_creation_failed():
         assert len(rsps.calls) == 0
 
 
+def test_api_mailboxes__same_local_part_on_different_domains():
+    """A domain admin should be able to create a mailbox with the same local part
+    of another mailbox, on different domain."""
+    # a mailbox exists on another domain
+    existing_mailbox = factories.MailboxFactory()
+
+    access = factories.MailDomainAccessFactory(
+        role=enums.MailDomainRoleChoices.ADMIN,
+        domain=factories.MailDomainEnabledFactory(),
+    )
+    client = APIClient()
+    client.force_login(access.user)
+
+    # create a full dict with same local part as another existing mailbox
+    mailbox_values = serializers.MailboxSerializer(
+        factories.MailboxFactory.build(local_part=existing_mailbox.local_part)
+    ).data
+
+    with responses.RequestsMock() as rsps:
+        rsps.add(
+            rsps.GET,
+            re.compile(r".*/token/"),
+            body='{"access_token": "domain_owner_token"}',
+            status=status.HTTP_200_OK,
+            content_type="application/json",
+        )
+        rsps.add(
+            rsps.POST,
+            re.compile(rf".*/domains/{access.domain.name}/mailboxes/"),
+            body=str(
+                {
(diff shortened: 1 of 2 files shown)


src/backend/mailbox_manager/tests/api/mailboxes/test_api_mailboxes_create.py
```diff
@@ -305,6 +305,37 @@ def test_api_mailboxes__create_pending_mailboxes(domain_status):
...
+    access = factories.MailDomainAccessFactory(
+        role=enums.MailDomainRoleChoices.ADMIN,
+        domain=factories.MailDomainEnabledFactory(),
+    )
+    existing_mailbox = factories.MailboxFactory(domain=access.domain)
+
+    client = APIClient()
+    client.force_login(access.user)
+    mailbox_values = dict(serializers.MailboxSerializer(existing_mailbox).data)
+    # import pdb; pdb.set_trace()
```
Morendil, line 321: Part of me is curious to know the story behind this line (and the one like it below). And part of me is afraid to hear it 😜

mjeammet: I just wanted to make sure all info were the same but only the `local_part` field should matter. I'll simplify it right away

src/backend/mailbox_manager/api/client/serializers.py
```diff
@@ -31,7 +31,12 @@ def create(self, validated_data):
         """
         Override create function to fire a request on mailbox creation.
         """
-        mailbox_status = enums.MailDomainStatusChoices.PENDING
+        if models.Mailbox.objects.filter(
+            local_part=validated_data["local_part"]
+        ).exists():
+            raise exceptions.ValidationError(
+                {"__all__": ["Mailbox with this Local_part and Domain already exists."]}
```
Morendil, line 38: This exception message mentions "local_part and domain" but the code actually only filters on local_part.

So if I have management access to more than one domain, it seems like trying to create a mailbox with the same local part as a mailbox that exists on *any* domain will raise this error ? The test doesn't seem to make this distinction.

I'm not very familiar with serializers so I might be missing something here.

src/backend/mailbox_manager/utils/dimail.py
```diff
@@ -68,8 +68,8 @@ def get_headers(self, user_sub=None):
 
         if response.status_code == status.HTTP_403_FORBIDDEN:
             logger.error(
-                "[DIMAIL] 403 Forbidden: Could not retrieve a token,\
-                please check 'MAIL_PROVISIONING_API_CREDENTIALS' setting.",
+                "[DIMAIL] 403 Forbidden: Could not retrieve a token,"
```
Morendil, line 71: Wow, I had no idea this was valid Python syntax. Learned something new today. (Also, apologies for taking two weeks to get around to a 2nd review of this PR…)

Morendil (approved):

src/backend/mailbox_manager/api/client/viewsets.py
```diff
@@ -230,6 +230,12 @@ def perform_create(self, serializer):
             )
         super().perform_create(serializer)
 
+    def get_serializer_context(self):
+        """Extra context provided to the serializer class."""
+        context = super().get_serializer_context()
+        context["domain_slug"] = self.kwargs["domain_slug"]
+        return context
+
```
sdemagny, line 238: pourquoi cet ajout ?

PR #540, opened 2024-11-20
🚚(backend) split api modules to improve readability

qbey:
## Purpose

The `core.api.viewsets` module and the `core.api.serializers` module are getting bigger: I propose to split them into submodules.

## Proposal

Use the https://pypi.org/project/gitfilesplit/ tool to split files without losing the history.

- [x] Split `core.api.viewsets`
- [x] Split `core.api.serializers`


sdemagny (approved):

Morendil:
Well that's my "Today I Learned" (TIL) - hadn't heard of GitFileSplit or this technique of using octopus merges. Thanks ;)

qbey:
Now is not the right time to do it, there is too much work conflicting in another PR...

PR #544, opened 2024-11-20
✨(dimail) automate allows requests to dimail when creating domain accesses

mjeammet:
## Purpose

Automatically send user creation and allow creation requests to dimail
upon creating domain access.

## Proposal

Description...

- [x] client function to send user creation request
- [x] client fuction to send allow creation request
- [x] test basic case for admin/owner
- [x] test basic case for viewer (no call made)
- [x] test wen user already exists
- [x] make sure it triggers when creating a domain !
- [ ] handle update of existing domain access + test OK when user goes from viewer to admin/owner (in another PR ?)

### DIFF
--- a/.github/workflows/people.yml
+++ b/.github/workflows/people.yml
@@ -197,6 +197,10 @@ jobs:
         run: |
           make demo FLUSH_ARGS='--no-input'
 
+      - name: Setup Dimail DB
+        run: |
+          make dimail-setup-db
+
       - name: Install Playwright Browsers
         run: cd src/frontend/apps/e2e && yarn install
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to
 
 ### Added
 
+- ✨(dimail) automate allows requests to dimail
 -  ✨(contacts) add notes & force full_name #565
 
 ### Changed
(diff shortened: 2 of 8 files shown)


mjeammet:
Something it breaking test `test_api_mail_domain__accesses_create_authenticated_administrator` : the `validate` method should raise an exception line 177 and prevent us from entering the `create`. Still, it seems like it doesn't stop us.

Do you have any idea what is happening, @sdemagny ?

mjeammet:
@qbey e2e tests are failing because of a navigation action that doesn't work as expected. Could it be due to the fact that, upon logging on localhost:3000 as people@people, I land on /teams/ and `mail-domains` tab icon has disappeared ? 
I feel like you mentioned something like that the other day.

mjeammet:
Reproduction en local de ce qui casse le test e2e : 
image

src/backend/core/api/client/viewsets.py
```diff
@@ -152,7 +152,7 @@ def list(self, request, *args, **kwargs):
             # - are profile contacts for a user
             user__isnull=True,
             # - are overriden base contacts
-            overridden_by__isnull=True,
+            overriding_contacts__isnull=True,
```
mjeammet, line 155: @Morendil c'est quoiiii çaaaaaa ?

Morendil (approved):

PR #560, opened 2024-11-26
✨(teams) add Team dependencies

qbey:
## Purpose

We want to allow Teams to be structured in trees to represent actual teams in a company.

## Proposal

Add the "tree" to the Team model and update the API endpoints to list parent teams along with the "direct" teams the user has access to.

This first implementation only allows to manually create those team trees. 
The API returns flat teams because we did not define the frontend behavior yet, this will be done later when we really know what we want.

The Treebeard has been pimped for PSQL use, and make the different values explicit.

- [x] add `django-treebeard` dependency
- [x] turn the `Team` model into an MP node
- [x] update APIs to list parent teams if the user belongs to the same organization only


src/backend/core/migrations/0010_team_depth_team_numchild_team_path_and_more.py
```diff
@@ -0,0 +1,55 @@
...
+from treebeard.mp_tree import MP_Node
+
+
+def update_team_paths(apps, schema_editor):
+    Team = apps.get_model("core", "Team")
+    alphabet = MP_Node.alphabet
+    steplen = MP_Node.steplen
+
+    # for default values, this resolves to: "{:04d}"
+    path_format = f"{{:{alphabet[0]}{steplen}d}}"
```
sampaccoud, line 14: I'm not sure this works because the alphabet contains letters :thinking: 
the path could be "1aZ6"

src/backend/core/models.py
```diff
@@ -633,7 +634,34 @@ def __str__(self):
...
+        if parent_id is None:
+            return self.model.add_root(**kwargs)
+
+        # Retrieve parent object, because django-treebeard uses raw queries for most
+        # write operations, and raw queries don’t update the django objects of the db
+        # entries they modify. See caveats in the django-treebeard documentation.
+        # This might be changed later if we never do any operation on the parent object
+        # before creating the child.
+        # Beware the N+1 here.
+        return self.get(pk=parent_id).add_child(**kwargs)
```
sampaccoud, line 661: why not expect the parent object directly? It seems to me what we expect when playing with objects.

src/backend/core/models.py
```diff
@@ -643,6 +671,12 @@ class Team(BaseModel):
     Team `service_providers`.
     """
 
+    # Allow up to 80 nested teams with 62^5 (916_132_832) root nodes
+    alphabet = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
+    # Django treebeard does not allow max_length = None...
+    steplen = 5
+    path = models.CharField(max_length=5 * 80, unique=True, db_collation="C")
```
sampaccoud, line 678: I was wondering if this collation impacts the use of "startswith" :thinking:

src/backend/demo/tests/test_commands_create_demo.py
```diff
@@ -1,29 +1,19 @@
...
-    "teams": 3,
-    "max_users_per_team": 5,
-    "domains": 2,
-}
-
 pytestmark = pytest.mark.django_db
 
 
 @override_settings(DEBUG=True)
-@mock.patch.dict(defaults.NB_OBJECTS, TEST_NB_OBJECTS)
```
sampaccoud, line 26: why not mock? It allows to run it with less objects in tests?

src/backend/demo/management/commands/create_demo.py
```diff
@@ -106,6 +133,20 @@ def __exit__(self, exc_type, exc_value, exc_tb):
...
+def _get_random_domain_name(excluded):
+    """Get a fake domain using an exclusion list."""
+    name = fake.domain_name()
+    for _max_retry in range(10):  # Limit iterations to avoid infinite loop
+        if name not in excluded:
+            return name
+        name = fake.domain_name()
+
+    if name in excluded:
+        raise RuntimeError("Could not find a domain name that is not already created")
```
sampaccoud, line 145: or add a iteration number aftter the domain? :man_shrugging:

src/backend/core/tests/test_models_teams.py
```diff
@@ -134,3 +134,85 @@ def test_models_teams_get_abilities_preset_role(django_assert_num_queries):
...
+    child_team = root_team.add_child(name="Child Team")
+    new_root_team = models.Team.add_root(name="New Root Team")
+    child_team.move(new_root_team, pos="first-child")
+    child_team.refresh_from_db()
+    assert child_team.get_parent(update=True) == new_root_team
+
+
+def test_models_teams_delete_team():
+    """
+    Delete a parent team also deletes children.
```
sampaccoud, line 180: in practice this will force you to introduce soft delete soon enough! :sweat_smile:

src/backend/core/models.py
qbey: > Using the C collation can improve database performance for certain operations, like sorting and comparison, because it uses simple byte-wise comparison rather than more complex locale-aware rules. However, this comes at the cost of losing locale-specific behaviors, such as case insensitivity and proper sorting for different languages.

So I guess it's only better :) Actually it still works properly in the specific case of "startswith" but I can't tell for the performance improvement or not.

src/backend/core/models.py
qbey: I don't use the object directly to be safer (even if sub-optimized) because as stated in django-treebeard documentation, it's important to have nodes up to date when doing such operations.

Do you think I should update the comment?

src/backend/core/tests/test_models_teams.py
qbey: Probably ^^ but for now this feature is hidden from end users, so I keep it for later when we need it.

src/backend/demo/tests/test_commands_create_demo.py
qbey: Yes, but the tests were not failing with the amount of objects in this tests, yet the command was failing. So I guess it was better to test the command as it is.

sampaccoud (approved):

PR #563, opened 2024-11-27
⚰️(contacts) remove "base" contact inheritance

qbey:
## Purpose

The base contact was initially made to allow users to override a contact which exists in a common imported directory. The need is still not very clear but it seems we will have another model dedicated to "imported contacts".

## Proposal

Remove all code related to `Contact.base`

- [x] Remove `Contact`'s `base` attribute
- [x] Remove related tests
- [x] Add migration


qbey:
We will keep the override possibility, this will be done in another PR.

PR #574, opened 2024-12-03
🔧(helm) fix siret validator for staging

sdemagny:
This fixes validator for the Organization
registration ID for staging (to enforce SIRET format).


qbey (approved):
Merci !

PR #577, opened 2024-12-03
Add siret attribute and mapper to Keycloak

Morendil:
## Purpose

Improve developer experience by making local dev resemble production conditions more closely.

## Proposal

We can now find organization data as provided by ProConnect in `user_info`, specifically SIRET.

### Add `OIDC_ORGANIZATION_REGISTRATION_ID_FIELD` to local dev settings

Otherwise local development would be blind to registration ID fields (e.g. SIRET)… and these will become more important as we start dealing with a variety of organizations both private and public (such as French communes).

### Expose organization registration ID field (e.g. SIRET) in /users/me endpoint

This is mostly to have a convenient way of testing the delivered feature, via an API-oriented E2E test. This isn't entirely satisfactory: as this is a very important feature, we should be able to unit test it. The E2E test depends on technical details such as the fact that we use Keycloak to simulate the ProConnect flow.

### Add `jq` to Python dev dependencies, so we can perform "surgery" on JSON responses

This is necessary for several backend tests that query the `/users` endpoint. These tests retrieve a JSON object and test for strict equality. But strict equality is much more stringent than what these tests intend to assert ! We only want to assert that queries for a given set of users will return all the users, and only those users, that the query designates. We are not interested in, for instance, the exact registration IDs of each of the users' organizations. (It is a matter for a more focused unit test to test, for instance, that the User serializer returns certain Organization fields.)

Also, if we tested against the actual, exact registration IDs of organizations, then our tests would become dependent on the order in which they are run, which is unacceptable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- 🧑‍💻(oidc) add ability to pull registration ID (e.g. SIRET) from OIDC #577
+
 ### Fixed
 
 - 🧑‍💻(user) fix the User.language infinite migration #611
--- a/docker/auth/realm.json
+++ b/docker/auth/realm.json
@@ -59,7 +59,7 @@
       "realmRoles": ["user"]
     },
     {
-      "username": "marie",
+      "username": "e2e.marie",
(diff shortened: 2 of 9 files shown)


qbey (approved):
nice :)

PR #583, opened 2024-12-05
Fetch and display commune name

### LINKED ISSUE #610: When an organization is created, the name fetched from the SIRET is always "Mairie".
## Bug Report

**Problematic behavior**
When an organization is created, the name fetched from the SIRET is always "Mairie".

See https://recherche-entreprises.api.gouv.fr/search?q=20010253100015

**Expected behavior/code**
In such case, we need to fetch a better name, maybe fetch the `nom_complet` when the name is "Mairie"?

**Steps to Reproduce**
1. In staging, login with a new user
2. Select a new "SIRET"
3. The new organization is created and the name is... "Mairie"

**Environment**
- People version: 1.9.0
- Platform:

**Possible Solution**
Update the `NameFromSiretOrganizationPlugin` to add a specific case

**Additional context/Screenshots**

image

Morendil:
## Purpose

New users from communes should be able to make sense of what they can do with Régie.

## Proposal

When arriving at Régie from ProConnect, make sure things are set up properly:
- lookup the Commune using the provided SIRET
- ensure the Organization is created with the commune's name

Fixes #610

### DIFF
--- a/src/backend/people/settings.py
+++ b/src/backend/people/settings.py
@@ -641,6 +641,8 @@ class Development(Base):
 
     OIDC_ORGANIZATION_REGISTRATION_ID_FIELD = "siret"
 
+    ORGANIZATION_PLUGINS = ["plugins.organizations.NameFromSiretOrganizationPlugin"]
+
     def __init__(self):
         """In dev, force installs needed for Swagger API."""
         # pylint: disable=invalid-name
@@ -685,6 +687,10 @@ class Test(Base):
 
     # this is a dev credentials for mail provisioning API
     MAIL_PROVISIONING_API_CREDENTIALS = "bGFfcmVnaWU6cGFzc3dvcmQ="
+
+    OIDC_ORGANIZATION_REGISTRATION_ID_FIELD = "siret"
+
+    ORGANIZATION_PLUGINS = ["plugins.organizations.NameFromSiretOrganizationPlugin"]
 
     ORGANIZATION_REGISTRATION_ID_VALIDATORS = [
         {
--- a/src/backend/plugins/organizations.py
+++ b/src/backend/plugins/organizations.py
@@ -3,6 +3,7 @@
 import logging
 
 import requests
+from requests.adapters import HTTPAdapter, Retry
 
 from core.plugins.base import BaseOrganizationPlugin
 
@@ -33,20 +34,17 @@ class NameFromSiretOrganizationPlugin(BaseOrganizationPlugin):
             logger.warning("Empty list 'liste_enseignes' in %s", organization_data)
         return None
 
-    def _get_organization_name_from_siret(self, siret):
-        """Return the organization name from the SIRET."""
-        try:
-            response = requests.get(self._api_url.format(siret=siret), timeout=10)
-            response.raise_for_status()
-            data = response.json()
-        except requests.RequestException as exc:
-            logger.exception("%s: Unable to fetch organization name from SIRET", exc)
-            return None
-
+    def get_organization_name_from_results(self, data, siret):
+        """Return the organization name from the results of a SIRET search."""
         for result in data["results"]:
+            nature = "nature_juridique"
+            commune = nature in result and result[nature] == "7210"
             for organization in result["matching_etablissements"]:
                 if organization.get("siret") == siret:
-                    return self._extract_name_from_organization_data(organization)
+                    if commune:
+                        return organization["libelle_commune"].title()
+
+                return self._extract_name_from_organization_data(organization)
 
         logger.warning("No organization name found for SIRET %s", siret)
         return None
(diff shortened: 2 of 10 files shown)


mjeammet (approved):
Nice ! Let's try and wait for Nathan for a few days if per chance he can adjust the design of the button by the end of the week. Otherwise, let's gooo !

Morendil:
@mjeammet This failed in CI for an interesting reason - the [API](https://recherche-entreprises.api.gouv.fr/) is rate-limited and we got a 429 error. I don't know how we managed to exceed 7 calls/sec. with only one E2E test exercising it… Not sure how best to proceed - perhaps implement retry logic, or try a different API, or use static data… it raises the question of how to deal with failure cases for something like this where the request is made at a specific point in the lifetime of an organization, and not retried afterward.

Morendil:
I have opted to implement retry logic, and merge with no further delay…

src/backend/people/settings.py
```diff
@@ -641,6 +641,8 @@ class Development(Base):
 
     OIDC_ORGANIZATION_REGISTRATION_ID_FIELD = "siret"
 
+    ORGANIZATION_PLUGINS = ["plugins.organizations.NameFromSiretOrganizationPlugin"]
+
```
qbey, line 645: This should not be enforced here because this is not generic.

src/backend/people/settings.py
```diff
@@ -686,6 +688,10 @@ class Test(Base):
     # this is a dev credentials for mail provisioning API
     MAIL_PROVISIONING_API_CREDENTIALS = "bGFfcmVnaWU6cGFzc3dvcmQ="
 
+    OIDC_ORGANIZATION_REGISTRATION_ID_FIELD = "siret"
+
+    ORGANIZATION_PLUGINS = ["plugins.organizations.NameFromSiretOrganizationPlugin"]
```
qbey, line 693: Same here, this should not be enforced here, only in specific tests.

src/frontend/apps/desk/src/features/header/AccountDropdown.tsx
```diff
@@ -14,7 +14,12 @@ export const AccountDropdown = () => {
     <DropButton
       button={
         <Box $flex $direction="row" $align="center">
-          <Text $theme="primary">{userName}</Text>
+          <Box $flex $direction="column" $align="left">
+            <Text $theme="primary">{userName}</Text>
+            {userData?.organization?.registration_id_list?.at(0) && (
```
qbey, line 19: I don't understand the condition, why should it be only when there is a registration ID?

src/helm/env.d/preprod/values.desk.yaml.gotmpl
```diff
@@ -55,7 +55,7 @@ backend:
     OIDC_RP_SCOPES: "openid email siret"
     OIDC_REDIRECT_ALLOWED_HOSTS: https://desk-preprod.beta.numerique.gouv.fr
     OIDC_AUTH_REQUEST_EXTRA_PARAMS: "{'acr_values': 'eidas1'}"
-    ORGANIZATION_PLUGINS: "plugins.organizations.NameFromSiretOrganizationPlugin"
+    ORGANIZATION_PLUGINS: ["plugins.organizations.NameFromSiretOrganizationPlugin"]
```
qbey, line 58: This will break, django-configurations does not expect a list here: https://github.com/jazzband/django-configurations/blob/master/configurations/values.py#L214

src/backend/plugins/organizations.py
```diff
@@ -33,20 +34,17 @@ def _extract_name_from_organization_data(organization_data):
...
         for result in data["results"]:
+            nature = "nature_juridique"
+            commune = nature in result and result[nature] == "7210"
             for organization in result["matching_etablissements"]:
                 if organization.get("siret") == siret:
-                    return self._extract_name_from_organization_data(organization)
+                    if commune:
+                        return organization["libelle_commune"].title()
+
+                return self._extract_name_from_organization_data(organization)
```
qbey, line 47: Warning, the indentation has changed here, this might get the wrong "siret".

src/backend/people/settings.py
Morendil: General reply to your comments: thanks for the feedbacks, I've already hit "merge" but can address these in a later PR (might pick #622 which is in some ways a continuation, or a new one altogether depending on urgency).

Specific reply to the above: OK, I understand the concern here is for users and contributors outside of France to be able to start from a vanilla configuration instead of having to figure out what is French-specific (or even French territorial administration specific) stuff that needs swapping out. And I don't think this case has arisen before so it's a good occasion to start documenting that.

On the other hand I'd like to keep these active for local development _and_ documented in the repo for ease of onboarding. Can we put these in, say, `env.d/development/france.dist` and pick them up in docker-compose.yml with a comment suggesting creating a country-specific set of env vars, as needed for other countries ?

src/frontend/apps/desk/src/features/header/AccountDropdown.tsx
Morendil: That does seem indirect now you mention it, testing for the name that we're going to display would work just as well. (I think I was biased by the testing perspective, the distinguishing feature of the organization that satisfies that test is that it's the only one in our test data that has a SIRET.)

src/helm/env.d/preprod/values.desk.yaml.gotmpl
Morendil: That one is confusing to me, originally I picked up the string value from this file and used it without change in `settings.py`, but that broke precisely because it was expecting a list of plugin names and not a singular name. (And it seemed reasonable to me that we'd want potentially more than one plugin…) Do we have different semantics between Helm files and `settings.py` ?

src/backend/plugins/organizations.py
Morendil: I'm not sure I see how to construct a case where we'd get the wrong result here (i.e. a missing unit test). The indent is by intention: I did mean to return directly with the `libelle_commune` in the commune case (because I'm pretty sure the key is always present) and fall through to the invocation of the private method otherwise. (So we end up with a bit of asymmetry in the code with some error handling in one case and not in the other.)

Morendil:
> I don't know how we managed to exceed 7 calls/sec. with only one E2E test exercising it…

Confirmed via Tchap by the Annuaire Entreprises team that the rate limit is global, not per IP or per user.

PR #588, opened 2024-12-06
💄(frontend) redirect home according to abilities

qbey:
## Purpose

We try to detect the landing page according to user permissions (abilities) instead of just the configuration setting.
This will be improved when the homepage is developed

## Proposal

- [x] Add abilities detection for root URL
- [x] Keep the default behavior if no abilities
- [x] Update jest tests
- [x] Update e2e tests

### DIFF
--- a/src/frontend/apps/desk/src/pages/index.tsx
+++ b/src/frontend/apps/desk/src/pages/index.tsx
@@ -1,16 +1,36 @@
 import { useRouter as useNavigate } from 'next/navigation';
 import { useEffect } from 'react';
 
+import { useAuthStore } from '@/core/auth';
 import { useConfigStore } from '@/core/config';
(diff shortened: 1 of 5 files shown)


src/frontend/apps/desk/src/pages/index.tsx
```diff
@@ -1,16 +1,36 @@
...
 const Page: NextPageWithLayout = () => {
   const { config } = useConfigStore();
+  const { userData } = useAuthStore();
   const router = useNavigate();
 
   useEffect(() => {
+    if (!userData) {
+      router.push('/authenticate/');
+      return;
+    }
```
qbey, line 17: The `useAuthStore` should already do that, not sure it's ok to do this again here, at least not this way?

PanchoutNathan (approved):

PR #592, opened 2024-12-09
🧑‍💻(dimail) allow account populate from DB

qbey:
## Purpose

Allow the dimail account creation command to create accounts from the data in people's database.

## Proposal

- [x] add a `--populate-from-people` argument to populate dimail's accounts

### DIFF
--- a/src/backend/mailbox_manager/management/commands/setup_dimail_db.py
+++ b/src/backend/mailbox_manager/management/commands/setup_dimail_db.py
@@ -8,7 +8,7 @@ import requests
 from rest_framework import status
 
 from mailbox_manager.enums import MailDomainStatusChoices
-from mailbox_manager.models import MailDomain
+from mailbox_manager.models import MailDomain, MailDomainAccess
 
 User = get_user_model()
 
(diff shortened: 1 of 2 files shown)


mjeammet:
Haven't reviewed it yet but I was thinking, now that my PR on dimail users + allows automatisation is about to be merged, that we could simply use core command "make demo" to populated dimail db. What do you think ?

qbey:
> Haven't reviewed it yet but I was thinking, now that my PR on dimail users + allows automation is about to be merged, that we could simply use core command "make demo" to populate dimail db. What do you think ? @qbey

Yes sure, but we need to keep commands separated to allow the run of this command "at will" to update from local data if someone created different accounts from the admin.

src/backend/mailbox_manager/management/commands/setup_dimail_db.py
```diff
@@ -157,3 +172,39 @@ def create_allows(self, user, domain, auth=(regie["username"], regie["password"]
...
+                domain.status = MailDomainStatusChoices.ENABLED
+                domain.save()
+            self.create_domain(domain.name)
+
+        # create missing users
+        for user in user_to_create:
+            self.create_user(
+                auth=(admin["username"], admin["password"]),
+                name=user.sub,
+                perms=["new_domain", "create_users", "manage_users"],  # give all perms
```
mjeammet, line 205: in dimail, we give no perms to basic users. Of the dimail users we're handling, only la_regie has these perms. 

I suggest reducing perms to [] to be iso with staging / preprod and prod environments.

src/backend/mailbox_manager/management/commands/setup_dimail_db.py
```diff
@@ -25,11 +25,23 @@ class Command(BaseCommand):
 
     help = "Populate local dimail database, for dev purposes."
 
+    def add_arguments(self, parser):
+        """Add arguments to the command."""
+        parser.add_argument(
+            "--populate-from-people",
```
mjeammet, line 31: How did you add this argument, locally ?
I think it would be neat to add a Make command (something like "dimail-populate-db").

mjeammet (approved):
Good to go after removal of excessive perms for basic users @qbey

PR #593, opened 2024-12-09
🧑‍💻(tilt) add dimail for local developments

qbey:
## Purpose

For local developments we add a dimail in the kubernetes stack

## Proposal

Description...

- [x] add kubernetes for the development stack in the kube configuration
- [x] update the start tilt script to looks like "meet" one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -354,9 +354,17 @@ start-kind: ## Create the kubernetes cluster
 	./bin/start-kind.sh
 .PHONY: start-kind
 
+install-external-secrets: ## install the kubernetes secrets from Vaultwarden
+	./bin/install-external-secrets.sh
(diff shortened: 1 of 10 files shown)


Morendil:
I've tried to make this work locally, unsuccessfully.

Unless I'm missing something, you need to be added to the recipients list in `secrets.enc.yaml` by someone who already has access, if you're a team member but do not have your identity already in `.sops.yaml` (which is my case). So, while useful to official team members, it seems that outside contributors won't be able to use Tilt out of the box.

Could the documentation in `docs/local_development_kube.md` make this distinction clearer and provide more guidelines for people who are joining (or already on) the team ?

For outside contributors, can we instead provide the development version of `secrets.enc.yaml` file in unencrypted form ?

PR #594, opened 2024-12-12
🐛(mailbox) fix activate mailbox feature

sdemagny:
Fix 422 Unprocessable Entity error returned
by dimail because of some missing data sent.


mjeammet (approved):
LGTM

PR #601, opened 2024-12-16
✨(dimail) check domain state

mjeammet:
## Purpose

Allows for request a domain check from dimail on a given domain where user is admin/owner. Dimail undergoes all verifications and returns domain state.

## Proposal

Description...

- [] item 1...
- [] item 2...

I suggest that 
if domain=="pending" and "status"=="ok": domain.status = enabled
if domain=="enabled" and "status"=="broken": domain.status = failed

### DIFF
--- a/src/backend/mailbox_manager/admin.py
+++ b/src/backend/mailbox_manager/admin.py
@@ -9,6 +9,30 @@ from mailbox_manager import models
 from mailbox_manager.utils.dimail import DimailAPIClient
 
 
+@admin.action(description=_("Check domain health"))
+def check_domain_health(modeladmin, request, queryset):  # pylint: disable=unused-argument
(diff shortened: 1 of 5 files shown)


src/backend/mailbox_manager/admin.py
```diff
@@ -9,6 +9,30 @@
...
+                request,
+                _(f"Check domain health failed for {domain.name} with message: [{err}]"),
+                )
+        else:
+            new_status = domain.status
+            if old_status != new_status:
+                domains_updated.append(domain)
+            messages.success(
+                request,
+                f"Domain updated {','.join(domains_updated)} {response.status_code} {response.json()}"
```
Morendil, line 32: This results in an exception when run locally, because `domains_updated` is a list of `MailDomain` objects rather than a list of `str` values…

PR #611, opened 2024-12-18
🧑‍💻(user) fix the `User.language`

qbey:
## Purpose

The use of a lazy function here make the Django migration detector to generate a migration every time we run `makemigrations`. 

## Proposal

This is not mandatory to have a lazy here as the settings are loaded once at runtime beginning.
As the choices makes noop migrations, we directly use the setting in the initial migration.

- [x] remove the lazy function
- [x] update the initial migration

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- 🧑‍💻(user) fix the User.language infinite migration #611
+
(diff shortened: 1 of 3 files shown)


mjeammet (approved):
Hallelujah !

No changelog ? Looks like a fix to me !

qbey:
> Hallelujah !
> 
> No changelog ? Looks like a fix to me !

It changes nothing actually ^^ but OK, I will add a changelog before merge :)

PR #612, opened 2024-12-19
🐛(dimail) fix mailboxes import from dimail

mjeammet:
## Purpose

I tried to import existing mailboxes from dimail for domain and raised an error because the response had an unexpected format.

update dimail response when requesting get_mailboxes

EDIT 06/01 : It seems to be fixed. I'm confused. I'll keep this PR to add tests regarding a new field in dimail's response : "additional Senders".

## Proposal

Description...

- [] item 1...
- [] item 2...

I'm getting bolossed by json. For an error that does not exist in prod, but I simply can't reproduce the behaviour in responsesMock. Send help, please.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to
 
 ## [Unreleased]
 
+- 🐛(dimail) fix mailboxes import from dimail #612
+
 ## [1.9.1] - 2024-12-18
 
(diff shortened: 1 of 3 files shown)


mjeammet:
can't reproduce the bug after cleaning of dimail's mailboxes in problematic domain.
let's close this PR and come back to it later, if needed.

PR #615, opened 2024-12-26
🧑‍💻(dimail) add missing John Doe access role on domain

sdemagny:


src/backend/mailbox_manager/management/commands/setup_dimail_db.py
```diff
@@ -81,6 +81,11 @@ def handle(self, *args, **options):
         else:
             self.create_user(name=people_base_user.sub)
             self.create_allows(people_base_user.sub, domain_name)
+            # access role for john doe is missing
+            maildomain = MailDomain.objects.get(name=domain_name)
+            MailDomainAccess.objects.create(
+                user=User.objects.get(email="people@people.world"),
+                role=enums.MailDomainRoleChoices.OWNER, domain=maildomain)
```
mjeammet, line 88: est-ce que si on utilisait la Factory, ça créera le "allow" aussi automatiquement ? Vraie question de retour de vacances ...

sdemagny: @mjeammet nope

mjeammet (approved):

src/backend/mailbox_manager/management/commands/setup_dimail_db.py
```diff
@@ -81,6 +81,13 @@ def handle(self, *args, **options):
         else:
             self.create_user(name=people_base_user.sub)
             self.create_allows(people_base_user.sub, domain_name)
+            # create access role for john doe
+            maildomain = MailDomain.objects.get(name=domain_name)
+            MailDomainAccess.objects.create(
+                user=User.objects.get(email="people@people.world"),
+                role=enums.MailDomainRoleChoices.OWNER,
+                domain=maildomain,
+            )
```
qbey, line 90: It should be under the `# we create a domain and add John Doe to it` comment? Because this part is more "people -> dimail"

Also, you should use a `get_or_create` to prevent fail when running several times.

PR #622, opened 2025-01-06
Automate domain creation

Morendil:
## Purpose

Completing #583, ensure users from communes can do something meaningful upon onboarding: create a default domain for them, so that they can create mailboxes

## Proposal

- Introduce ApiCall object, to avoid mocking and anticipating task queuing
- Call Scaleway API upon organization creation to provision subzone and email-related DNS entries

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,12 @@ and this project adheres to
 
 ### Added
 
+- ✨(anct) auto create domains #622
 - ✨(teams) add Team dependencies #560
 - ✨(organization) add admin action for plugin #640
 - ✨(anct) fetch and display organization names of communes #583
 - ✨(frontend) display email if no username #562
+- ✨(anct) fetch and display organization names of communes #583
 - 🧑‍💻(oidc) add ability to pull registration ID (e.g. SIRET) from OIDC #577
 
 ### Fixed
--- a/src/backend/core/tests/resource_server_api/teams/test_list.py
+++ b/src/backend/core/tests/resource_server_api/teams/test_list.py
@@ -192,9 +192,9 @@ def test_api_teams_order_param(client, force_login_via_resource_server):
     response_data = response.json()
     response_team_ids = [team["id"] for team in response_data["results"]]
 
-    assert (
-        response_team_ids == team_ids
-    ), "created_at values are not sorted from oldest to newest"
+    assert response_team_ids == team_ids, (
+        "created_at values are not sorted from oldest to newest"
+    )
 
 
 def test_api_teams_list_with_parent_teams(client, force_login_via_resource_server):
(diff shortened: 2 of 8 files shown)


Morendil:
@qbey Let me know when we could discuss this, even in draft status: this takes advantage of the plugin idea but aims it in a somewhat different direction, perhaps, than what your original PR implied.

I'm thinking that the plugins system could be a nice place to locate the responsibility for business rules such as attributing the admin role to a user automatically (what I have in mind for the communes use case, but that may not generalize to other kinds of orgs). For this to work we'd have to introduce additional "seams" corresponding to further points in the lifecycle, such as upon attributing the admin access.

qbey (commented):
@Morendil 

The plugin system was designed for this, yes :) This first implementation does not respect the concern of genericity:

- `NameFromSiretOrganizationPlugin` must not be replaced, the `CommuneCreation` must be another plugin.
- We may add a field to the Organization like `extra`  which would be a generic JSON field to store if the organization is a "commune", information we can get from `NameFromSiretOrganizationPlugin` with few modifications.
- Then the `CommuneCreation` would check if the Organization is a "commune", in which case it could trigger the domain creation (and maybe other stuffs).

I don't like the `ApiCall` object, it adds an overhead on something already simple. If you like, we can think about a small "Request" override to simplify the "retry", but I would prefer to keep this for later.

Morendil:
@qbey Thanks for the feedback, only getting back to it now after recovering from a bout of the flu…

I'm working on a "take 2", which you can see at branch `lbo/automate-domain-creation-take2` (which I will overwrite this PR's contents with if it seems suitable).

As a preliminary remark I take your comment on "does not respect the concern of genericity" as something that is a constraint for future functionality. I understand, informally, the reasons for such a constraint - because we've discussed some of them at the lunch table - but this isn't something that, to my knowledge, has been documented and made explicit previously, so I would like to suggest writing up this topic if only to ease onboarding of future devs.

I've rolled back the changes to NameFromSiret that made a special case for communes, because it was evident that this resulted in much more awkward code, and it seems that the CommuneCreation plugin is better suited to take over the specific case of communes.

However, this gives rise to a new problem: now both plugins have to make a call to the Annuaire API, increasing the risk of bumping against rate limits, and resulting in code duplication.

Would it be acceptable to, as you suggested, store the entire Annuaire API results in a JSON field ? So the NameFromSiret would become instead OrganizationDataFromSiret. Then this allows for later plugins to exploit this information in different ways.

As for the approach to testing and implementing the (somewhat complex) dance of API calls needed for domain creation, I ask that you trust that I know what I'm doing.

I know that this is not the way things are done so far in this code base, but I'm confident in the ability of our devs to pick up new techniques, and I believe we have sufficient evidence of pain points around testing API calls to external services (Dimail in particular), and I also have a plan in mind for how this approach can help, as well as unlock interesting new capabilities.

I believe (though anyone is free to disagree) that my track record of valuable contributions so far to this code base is a reasonable argument for requesting this vote of confidence.

Morendil:
Closing in favor of a new version "take 2"

PR #634, opened 2025-01-10
✨(frontend) display email if no username

### LINKED ISSUE #562: Si pas de name pour le user afficher l'email dans la partie mon compte

sdemagny:
Fallback to email if no username is defined for logged user.
close #562

### DIFF
--- a/src/frontend/apps/desk/src/features/header/AccountDropdown.tsx
+++ b/src/frontend/apps/desk/src/features/header/AccountDropdown.tsx
@@ -9,7 +9,7 @@ export const AccountDropdown = () => {
   const { t } = useTranslation();
   const { userData, logout } = useAuthStore();
 
-  const userName = userData?.name || t('No Username');
+  const userName = userData?.name || userData?.email || t('No Username');
   return (
     <DropButton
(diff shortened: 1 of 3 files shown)


mjeammet (approved):

src/frontend/apps/desk/src/features/header/AccountDropdown.tsx
```diff
@@ -9,7 +9,7 @@ export const AccountDropdown = () => {
   const { t } = useTranslation();
   const { userData, logout } = useAuthStore();
 
-  const userName = userData?.name || t('No Username');
+  const userName = userData?.name || userData?.email || t('No Username');
```
Morendil, line 12:
```suggestion
  const userName = userData?.name || userData?.email;
```

`User.email` is not nullable, so the last alternative no longer needs to appear.

sdemagny: @Morendil User.email can be null
--> https://github.com/numerique-gouv/people/blob/main/src/backend/core/models.py#L439

PR #638, opened 2025-01-14
✨(oidc) people as an identity provider

qbey:
## Purpose

We want the people user database to also be an Identity Provider.

Ref: 

- https://github.com/numerique-gouv/proconnect-documentation/blob/main/doc_fi/README.md
- https://github.com/numerique-gouv/proconnect-documentation/blob/main/doc_fi/configuration.md

## Proposal

This is the very first step toward this goal:

- [x] Allow people to ask like an OIDC Identity Provider
- [x] Add a login page to allow login through email/password
- [x] Provide a local testing environment
- [x] Add cooldown when successive login fail
- [x] Add a check on `acr` to refuse higher level than `eidas1` when requested by ProConnect (comes from the SP)
- [x] Use the `login_hint` for the login page (and don't allow email field edition?)
- [x] Do not login the user on people if the OIDC authentication loop fails (aka do not create a session when login from the frontend login page)
- [x] Add backend tests
- [ ] Request help to clean the frontend part ^^
- [x] Add frontend tests
- [x] Add E2E tests


jblagadec:
**COMMENTAIRE ORAL QUENTIN**
Boîte mail permet de s'authentifier sur la Régie. Donne accès, couplé avec un mdp.
En passant par Proconnect : crée un user Django.

qbey:
Changelog will be added when review is done (to ease the rebase process...)

src/backend/mailbox_manager/models.py
```diff
@@ -23,6 +24,13 @@ class MailDomain(BaseModel):
     name = models.CharField(
         _("name"), max_length=150, null=False, blank=False, unique=True
     )
+    organization = models.ForeignKey(
+        Organization,
+        on_delete=models.PROTECT,
+        related_name="mail_domains",
+        null=True,
```
sylvinus, line 31: Is the plan to later switch this to null=False ?

src/backend/mailbox_manager/utils/dimail.py
```diff
@@ -341,6 +342,7 @@ def import_mailboxes(self, domain):
                             # secondary email is mandatory. Unfortunately, dimail doesn't
                             # store any. We temporarily give current email as secondary email.
                             status=enums.MailboxStatusChoices.ENABLED,
+                            password=make_password(None),
```
sylvinus, line 345: This could use a comment. Are we absolutely sure it can't lead to successful with empty passwords?

src/backend/mailbox_manager/api/client/serializers.py
```diff
@@ -34,7 +36,7 @@ def create(self, validated_data):
         """
         Override create function to fire a request on mailbox creation.
         """
-        mailbox = super().create(validated_data)
+        mailbox = super().create(validated_data | {"password": make_password(None)})
```
sylvinus, line 39: same feedback here

src/backend/mailbox_manager/models.py
```diff
@@ -202,6 +218,9 @@ class Mailbox(BaseModel):
         choices=MailboxStatusChoices.choices,
         default=MailboxStatusChoices.PENDING,
     )
+    dn_email = models.EmailField(_("email"), blank=True, unique=True, editable=False)
```
sylvinus, line 221: what is dn_ ? isn't it a potential issue to store both this and the local_part ?

src/backend/mailbox_manager/utils/dimail.py
qbey: Yes, when `None` it starts with the "invalid password" prefix.

See https://github.com/django/django/blob/5.1.6/django/contrib/auth/hashers.py#L103-L106
and https://github.com/django/django/blob/5.1.6/django/contrib/auth/hashers.py#L28-L42

I will add a comment. I don't like testing the framework but I can test "None" password cannot login, just in case.

src/backend/mailbox_oauth2/tests/test_backends.py
```diff
@@ -0,0 +1,258 @@
...
+    """Test get_user with inactive user."""
+    organization = core_factories.OrganizationFactory(with_registration_id=True)
+    domain = factories.MailDomainEnabledFactory(organization=organization)
+    mailbox = factories.MailboxFactory(domain=domain)
+
+    user = MailboxModelBackend().get_user(mailbox.pk)
+    assert user is None
+
+
+def test_get_username_domain_from_email_valid():
```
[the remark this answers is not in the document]
qbey: Definitely :+1:

src/backend/mailbox_manager/models.py
qbey: A domain doesn't really require an Organization today, it's only mandatory for ProConnect for now... But when we will improve permissions, it may become:

- easy to determine the organization, or at least propose a set of organization to the end user
- mandatory to have an organization to be able to determine who can manage mailboxes

So yeah, it may become null=False in a near future :) While the "near future" is not clear, I did not mention it in the code ^^'

src/backend/mailbox_manager/models.py
qbey: "dn_" stands for "denormalized". In the first place, I tried to add a generated field but it does not work with "join" (it was a concat of "local_part", "@" and "domain__name" :boom: ), so I used a field which is computed on save to store the denormalized email value. 

For sure, if you make `Mailbox.objects.filter().update(local_part="blah")`, the `dn_email` will be broken... 

The `dn_email` is required to make the Django admin work for mailbox with password...

I can add a comment to say we should only use this field for the Django admin.

qbey (commented):
@sylvinus Thanks for the review, I will improve that :)

src/backend/mailbox_oauth2/backends.py
```diff
@@ -0,0 +1,125 @@
...
+            logger.error(
+                "EmailModelBackend was triggered but the `one_time_email_authenticated_session` "
+                "is not set: ignoring authentication."
+            )
+            return None
+
+        # Check if the account is locked
+        if self._check_login_attempts(email):
+            logger.warning("Account locked due to too many failed attempts: %s", email)
+            Mailbox().set_password(password)
```
sylvinus, line 90: could use the same comment as below (moreover, it's the first). great technique btw!

src/backend/mailbox_oauth2/tests/test_middleware.py
```diff
@@ -0,0 +1,150 @@
...
+    )
+    mailbox = factories.MailboxEnabledFactory(domain=domain, local_part="user")
+
+    response = client.post(
+        "/api/v1.0/login/", {"email": "user@example.com", "password": "password"}
+    )
+    assert response.status_code == 200
+
+    # assert the user has a session
+    assert not client.session.is_empty()
```
sylvinus, line 82: Random thought, is there an expiration time on this session? it should always be short-lived, right?

sylvinus:
Great patch congrats!

Given the sensitive nature of providing OIDC identities, I think we need to do a careful security review.

For instance, we should look into security headers on the login page (https://www.cerberauth.com/blog/openid-connect-provider-security-headers/) Potential list (for next and/or django?)
```
{
    'X-Frame-Options': 'DENY',
    'X-Content-Type-Options': 'nosniff',
    'X-XSS-Protection': '1; mode=block',
    'Strict-Transport-Security': 'max-age=31536000; includeSubDomains',
    'Content-Security-Policy': "default-src 'self'; frame-ancestors 'none'",
    'Referrer-Policy': 'strict-origin-when-cross-origin'
}
```

src/backend/mailbox_manager/models.py
sylvinus: Could this be a reason to do the reverse and compute local_part while only storing the full email?

qbey:
@sylvinus In production, the Django headers are already quite well set, see for instance on /api/v1.0/authenticate/:

```
{
    'Date': 'Wed, 12 Feb 2025 23:25:20 GMT', 
    'Content-Type': 'text/html; charset=utf-8', 
    'Content-Length': '0',
    'Connection': 'keep-alive', 
    'Allow': 'GET', 
    'Vary': 'origin, Accept-Language, Cookie', 
    'X-Frame-Options': 'DENY',
    'Content-Language': 'en-us', 
    'Strict-Transport-Security': 'max-age=31536000; includeSubDomains',
    'X-Content-Type-Options': 'nosniff', 
    'Referrer-Policy': 'same-origin', 
    'Cross-Origin-Opener-Policy': 'same-origin',
}
```

src/backend/mailbox_oauth2/tests/test_middleware.py
qbey: You are right, even if the session cannot be used in any other context, I reduced it to one minute.

src/backend/mailbox_manager/models.py
qbey: We could, but it implies a bit of work, if we need it we may do it in another PR if it's ok for you.

lunika (commented):
I lack a lot of context to do a relevant code review I think.
I mostly focus on the "code" itself more than its purpose on the global project.

src/backend/mailbox_oauth2/serializers.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+    password = serializers.CharField(
+        write_only=True,
+        help_text="User's password for authentication",
+        required=True,
+    )
+
+    def create(self, validated_data):
+        """Do not allow creating instances from this serializer."""
+        raise RuntimeError("LoginSerializer does not support create method")
```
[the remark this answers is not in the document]
qbey: Because it should not be used

src/backend/people/settings.py
```diff
@@ -682,6 +682,24 @@ def post_setup(cls):
...
+        )
+
+        # - Serialize private key to PEM format
+        private_key_pem = private_key.private_bytes(
+            encoding=serialization.Encoding.PEM,
+            format=serialization.PrivateFormat.TraditionalOpenSSL,
+            encryption_algorithm=serialization.NoEncryption(),
+        )
+
+        return private_key_pem.decode("utf-8")
```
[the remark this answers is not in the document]
qbey: It's used in `class Local(Production)` that's why I've put it in the Base class. For sure it can be used somewhere we don't want, but in the end it won't create any security issue, so I will leave it here.

PR #641, opened 2025-01-16
🐛(frontend) improve e2e tests avoiding race condition from mocks

Morendil:
## Purpose

Several E2E tests were suffering from a race condition problem. The race condition affected setting up mocked API calls with `page.route` and navigation to the front page of Desk. Specifically, the call to `page.goto('/')` was made _before_ the call to `page.route`, without caring to synchronize the two. This meant that, at random, we could possibly have the API calls made by the front page complete before the mocks were in place, leading to a test failure.

## Proposal

This PR aims to eliminate this source of intermittent CI failures, which are always obnoxious and lead to a waste of time (when we investigate why the build failed), of resources (when we have to rerun jobs) and more importantly a loss of confidence in the test suite.

We systematically have the calls to `page.route` take place (synchronously, with an `await`) before calling `page.goto('/')`. (Unfortunately this means that we cannot make use of `test.beforeEach` in these tests, a price I'm happy to pay as these clauses were only a couple lines each).

(There is one exception to this "rule", `menu-spec.ts`, because the login flow is triggered by a call to the `/users/me` endpoint, which must return a 401 on the first invocation; setting up the mock before the call to `page.goto('/')` interferes with that. We must trigger the login process by loading the home page, then prepare the mock, then proceed with login. Unfortunately I'm not able to find a fix for this that works both locally and in CI, so I'm leaving `menu-spec.ts` unchanged for now.)

In addition, fix some E2E calls that were not making use of `await`, to be more in line with all other E2E tests.

**In addition, lower the default timeout for Playwright tests to 10 seconds, which is plenty.**

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-desk/mail-domains.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-desk/mail-domains.spec.ts
@@ -73,16 +73,14 @@ const mailDomainsFixtures: MailDomain[] = [
 
 test.describe('Mail domains', () => {
   test.describe('checks all the elements are visible', () => {
-    test.beforeEach(async ({ page, browserName }) => {
+    test('checks the sort button', async ({ page, browserName }) => {
       await page.goto('/');
       // The user is a team administrator, so they land on the team page
       // This allows to prevent the redirection to the mail domains page
       // to make the '/api/v1.0/mail-domains/?page=1&ordering=created_at'
       // query at login, which will be cached and not called afterward in tests.
       await keyCloakSignIn(page, browserName, 'team-administrator-mail-member');
-    });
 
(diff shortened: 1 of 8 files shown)


src/frontend/apps/e2e/__tests__/app-desk/mail-domains.spec.ts
```diff
@@ -164,6 +166,10 @@ test.describe('Mail domains', () => {
         });
       });
 
+      await page.goto('/');
+      // See comment above
```
sdemagny, line 170: vaudrait mieux être explicite à chaque fois non ?

Morendil: ✅

elvoisin (approved):

PR #658, opened 2025-01-22
Automate domain creation

Morendil:
## Purpose

Completing #583, ensure users from communes can do something meaningful upon onboarding: create a default domain for them, so that they can create mailboxes

## Proposal

- Segregate France-specific dev settings so that they are easy to disable for other countries
- Call Scaleway API upon organization creation to provision subzone and email-related DNS entries
- Extend plugin mechanism to grant admin access to the newly created domain in Dimail
- Introduce ApiCall object, to avoid mocking and anticipating task queuing

### DIFF
--- a/src/backend/core/plugins/base.py
+++ b/src/backend/core/plugins/base.py
@@ -11,3 +11,9 @@ class BaseOrganizationPlugin:
     def run_after_create(self, organization) -> None:
         """Method called after creating an organization."""
         raise NotImplementedError("Plugins must implement the run_after_create method")
+
+    def run_after_grant_access(self, organization_access) -> None:
+        """Method called after creating an organization."""
+        raise NotImplementedError(
+            "Plugins must implement the run_after_grant_access method"
+        )
--- a/src/backend/plugins/organizations.py
+++ b/src/backend/plugins/organizations.py
@@ -1,8 +1,10 @@
 """Organization related plugins."""
 
 import logging
+import re
 
 from django.conf import settings
+from django.utils.text import slugify
 
 import requests
 from requests.adapters import HTTPAdapter, Retry
@@ -139,6 +141,8 @@ class CommuneCreation(BaseOrganizationPlugin):
 
     def dns_call(self, spec):
         """Call to add a DNS record"""
+        zone_name = self.zone_name(spec.inputs["name"])
+
         records = [
             {
                 "name": item["target"],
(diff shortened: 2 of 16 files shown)


src/backend/core/plugins/base.py
```diff
@@ -11,3 +11,9 @@ class BaseOrganizationPlugin:
     def run_after_create(self, organization) -> None:
         """Method called after creating an organization."""
         raise NotImplementedError("Plugins must implement the run_after_create method")
+
+    def run_after_grant_access(self, organization_access) -> None:
+        """Method called after creating an organization."""
+        raise NotImplementedError(
+            "Plugins must implement the run_after_grant_access method"
```
sylvinus, line 18: Should these plugin methods really be mandatory?

src/backend/plugins/organizations.py
```diff
@@ -81,3 +81,198 @@ def run_after_create(self, organization):
...
+        logger.info(
+            "API call: %s %s %s %s",
+            self.method,
+            self.url,
+            self.params,
+            self.response_data,
+        )
+
+
+class CommuneCreation(BaseOrganizationPlugin):
```
sylvinus, line 128: As this is French-specific, this code probably shouldn't be in a `plugins/organizations.py` file ? Should we have a "fr" folder there?

sylvinus (commented):
Here we are doing sequential requests on different APIs, what happens if any of them fail along the way? i.e. what state will our organization be in afterwards?

src/backend/core/plugins/base.py
Morendil: In this PR the intent is to extend the plugin mechanism (introduced in #584) remaining consistent with prior choices, among which was this abstract base class behaviour. This is common in Python and could have other rationales, such as protecting against non-intuitive method dispatching, or ensuring plugin implementors take care to consider the entire life cycle when creating a plugin.

We can loop in @qbey ?

src/backend/plugins/organizations.py
Morendil: Similar reasoning to the above: my intent was to extend the existing plugin mechanism and not break new ground.

Morendil:
> Here we are doing sequential requests on different APIs, what happens if any of them fail along the way? i.e. what state will our organization be in afterwards?

The Organization object will not be much affected by failures, in general. We only update the name.

If the call to fetch the name fails, then the Organization is stuck with having its SIRET as its name, since the plugin is only called on creating the Organization object. (This is one reason I'm thinking we might need an asynchronous mechanism that lets us specify tasks and retry them later if they fail at this stage.)

If the calls to the Scaleway API fail, then the Organization is in a usable state (users can create mailboxes, etc.) but users will not be able to actually send or receive emails. It's not too hard to fix an occasional failure by hand, but if they are more common in practice then again an asynchronous queue where tasks can be retried could be a good idea.

Same sort of situation if the calls to Dimail fail - the internal state of Régie is still consistent and users will still be able to create mailboxes, but we would have to rectify the situation by hand at some point. There is already some code that will ask Dimail to perform the mailbox creations if the creation of the mail domain failed.

So, overall there are already some measures in place making the system resilient, and I think it's worth seeing what happens in practice before deciding whether we need to think about architectural support for making this resilience more systematic.

qbey (approved):

PR #659, opened 2025-01-22
🐛(dimail) fix imported mailboxes should be enabled instead of pending

mjeammet:
## Purpose

Importing mailboxes creates pending mailboxes ... but these mailboxes are already active and, for most, functional. We thus mark them as "enabled".

## Proposal

Description...

- [x] fixed related test
- [x] set imported mailboxes' statuses to "enabled"
- [ ] for future PR : dimail provides a "status" field that can be either "ok" or "broken". our mailboxes could depend on this status but it's currently not entirely reliable (many "broken" mailboxes actually are functioning).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


sdemagny (approved):
;)

PR #660, opened 2025-01-22
⚡️(api) add cache to prevent excessive drain from /stats endpoint

mjeammet:
## Purpose

stats endpoint is currently public and without cache. As it makes several queries to the database, we want to add a cache to prevent excessive drain

## Proposal

Description...

- [x] added cache
- [x] added a test, out of curiosity


sylvinus (approved):
LGTM

PR #662, opened 2025-01-25
✨(mailbox_manager) add shedule task to fetch domains status

### LINKED ISSUE #695: Deploiement celery pour tache planifié #670

sdemagny:
Add celery crontab to check and update domains status. This task calls dimail API.

TODO: 

- [x] improve task return to give more information
- [ ] add flower to helm chart => not here, needs a dedicated issue in best effort
- [x] bump version in Chart.yaml
- [ ] deploy new helm chart using github action

FIXES #695

### DIFF
--- a/src/backend/mailbox_manager/tasks.py
+++ b/src/backend/mailbox_manager/tasks.py
@@ -1,0 +1,65 @@
+"""Mailbox manager tasks."""
+
+import time
+
+import requests
+from celery import Celery
+from celery.schedules import crontab
+from celery.utils.log import get_task_logger
+
+from mailbox_manager import enums
+from mailbox_manager.models import MailDomain
+from mailbox_manager.utils.dimail import DimailAPIClient
+from people.celery_app import app as celery_app
+
+logger = get_task_logger(__name__)
+
+
+@celery_app.on_after_finalize.connect
+def setup_periodic_tasks(sender: Celery, **kwargs):
+    """Setup periodic tasks."""
+    sender.add_periodic_task(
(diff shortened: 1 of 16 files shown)


qbey (commented):
Did you consider adding the sentry for celery? https://docs.sentry.io/platforms/python/integrations/celery/

Morendil (commented):
When trying this locally, a file `src/backend/celerybeat-schedule` is created - it would be good to add it to `.gitignore` ?

src/backend/mailbox_manager/tasks.py
```diff
@@ -0,0 +1,54 @@
...
+            client.fetch_domain_status(domain)
+        except requests.exceptions.HTTPError as err:
+            logger.error("Failed to fetch status for domain %s: %s", domain.name, err)
+        else:
+            if old_status != domain.status:
+                update_count += 1
+            else:
+                check_count += 1
+    # todo: add more details to store more information in the database
+    return f"Domains processed: {update_count} updated, {check_count} checked"
```
Morendil, line 46: The counting logic is counter-intuitive. A domain is counted as _either_ checked or updated (so you can have 1 updated and 0 checked). What I would expect is that "updated" is a subset of "checked"…

Also, domains that failed verification are not counted at all… and we probably want to report them because the failure logs are going to be hidden in the Celery container's logs, not directly visible here.

I would prefer something like

`f"Domains processed: {same_count} not changed, {update_count} updated, {failure_count} failed, {check_count} checked"`

with the invariant same_count+update_count+failure_count == check_count.

PR #675, opened 2025-01-30
fix doc and add age key

sdemagny:


qbey (approved):

mjeammet (changes requested):
Jacques confirmed we don't use age keys anymore with the new system

PR #676, opened 2025-01-30
🚑️(teams) do not display add button when disallowed

qbey:
## Purpose

The frontend should not display "add team" actions when the user has not the ability.
When we don't allow the user to see the team creation button, we also want to disable the corresponding API.

## Proposal

- [x] hide team creation button according to user abilities
- [x] check user abilities on team creation endpoint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to
 
 ### Fixed
 
+- 🚑️(teams) do not display add button when disallowed #676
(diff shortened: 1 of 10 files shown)


sdemagny (approved):
Juste dans le message de commit remets le "s" de display stp -> [🚑️(teams) do not diplay add button when disallowed](https://github.com/suitenumerique/people/pull/676/commits/fbc246e6b6e71d777070a86e51a80c019bdf110a)

jblagadec:
Pas de recette possible sans voir la section Groupe

PR #687, opened 2025-02-03
✨(auth) fix empty name from ProConnect

mjeammet:
## Purpose

user names are supposed to be computer automatically at user creation This commit fixes the way names are computed from ProConnect claims

## Proposal

Description...

- [] item 1...
- [] item 2...


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to
 
 ### Fixed
 
+-  ✨(auth) fix empty names from ProConnect #687
```
qbey, line 19:
```suggestion
- ✨(auth) fix empty names from ProConnect #687
```

qbey (commented):
It there a test to add to mimic what happens in production?

CHANGELOG.md
mjeammet: We have a dedicated test (test_authentication_getter_new_user_with_email) but it was not broken prior to this PR as prod was. I'm not confident the way we mock is the way ProConnect sends its data. I'm going to check that out.

qbey:
This won't fix the problem. The issue in production environment is the setting defining the fields from which the name is built is not defined.

src/backend/core/authentication/backends.py
```diff
@@ -220,7 +219,7 @@ def compute_full_name(self, user_info):
         """Compute user's full name based on OIDC fields in settings."""
         name_fields = settings.USER_OIDC_FIELDS_TO_NAME
         full_name = " ".join(
-            user_info[field] for field in name_fields if user_info.get(field)
+            user_info.get(field) for field in name_fields if user_info.get(field)
```
qbey, line 222: This change does nothing because `user_info[field]` is always ok when `user_info.get(field)` is truthy.

src/helm/env.d/dev/values.desk.yaml.gotmpl
```diff
@@ -47,7 +47,10 @@ backend:
         name: backend
         key: OIDC_RP_CLIENT_SECRET
     OIDC_RP_SIGN_ALGO: RS256
-    OIDC_RP_SCOPES: "openid email siret"
+    OIDC_RP_SCOPES: "openid email siret given_name usual_name"
+    USER_OIDC_FIRST_NAME_FIELD: "given_name"
+    USER_OIDC_LAST_NAME_FIELD: "usual_name"
```
qbey, line 52: Please don't :) Those settings will be added in a future PR

src/helm/env.d/dev/values.desk.yaml.gotmpl
```diff
@@ -47,7 +47,10 @@ backend:
         name: backend
         key: OIDC_RP_CLIENT_SECRET
     OIDC_RP_SIGN_ALGO: RS256
-    OIDC_RP_SCOPES: "openid email siret"
+    OIDC_RP_SCOPES: "openid email siret given_name usual_name"
```
qbey, line 50: Scopes don't work like that
```suggestion
    OIDC_RP_SCOPES: "openid email siret profile"
```

src/helm/env.d/dev/values.desk.yaml.gotmpl
mjeammet: I can't crack WHY but it stops working when I remove it. And of course I restarted tilt, and tried on and off and on again and I definitely need them. 

I use the `tilt-up` stack, no keycloak

mjeammet: I know it doesn't make any sense and I can't crack WHY but name retrieval stops working when I remove it. Of course I restart tilt eveytime I try, and tried on and off and on again and it definitely seem to need them. 
.
I use the `tilt-up` stack by the way, no keycloak

src/helm/env.d/dev/values.desk.yaml.gotmpl
```diff
@@ -47,7 +47,8 @@ backend:
...
```
mjeammet, line 50: @qbey I'm unsure whether I should put the same scope on the keycloak config or if it doesn't matter because keycloak doesn't use these scopes anyway.

qbey: Keycloak uses the standard : openid email profile siret

qbey (approved):

PR #689, opened 2025-02-04
📝(tilt) improve tilt doc

mjeammet:
## Purpose

I'm trying to install tilt and doc seems outdated/Mac-only. This is an attempt at making it better, especially for linux users.

## Proposal

Description...

- [x] upgrade tilt doc by remove sops-related topics
- [x] how to add secrets

### DIFF
--- a/docs/local_development_kube.md
+++ b/docs/local_development_kube.md
@@ -5,6 +5,8 @@ Tilt is a tool that helps you develop applications for Kubernetes.
 It watches your files for changes, rebuilds your containers, and restarts your pods. 
 It's like having a conversation with your cluster.
 
+This is particularly useful when working on integrations or to test your helm charts.
+Otherwise, you can use your good old docker configuration as described in README.md. 
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


docs/local_development_kube.md
```diff
@@ -10,116 +10,48 @@ It's like having a conversation with your cluster.
...
 
 ```bash
 make install-external-secrets
 ```
 
 ### Deploy the application
 
-```bash
-# Pro Connect environment
-tilt up -f ./bin/Tiltfile
```
qbey, line 104: This is the command I use, why remove it? It's common for Tilt users to use it like this

docs/local_development_kube.md
```diff
@@ -10,116 +10,48 @@ It's like having a conversation with your cluster.
 
 This guide assumes you have the following tools installed:
 
+- [brew](https://brew.sh/) if you are on linux
```
qbey, line 13: I did not need brew on Linux. Do you mean MacOS?
I was hoping someone on MacOS could write a dedicated part about the setup.

qbey (commented):
Thanks for the cleanup, I'm not fully convinced we should remove the "command without make" part but if you think it helps go ahead :)

docs/local_development_kube.md
mjeammet: but since you made the `make tilt-up` command, I thought it would lighten the doc to not put both commands everytime. But I lack experience wiht tilt. If this command is especially familiar with the community, maybe we should keep this command and add `make tilt-up` in parenthesis or something like that.

docs/local_development_kube.md
mjeammet: Didn't you ? Well, I'm on Ubuntu and I there is at least one of these tools (helmfile, I think) is not Debian-friendly at all. Do you remember how you installed it ? Or are you running it as a container ?

I believe @Morendil is also trying to set this up on Mac. Maybe he can pitch in.

qbey: The helmfile I used (in docker) is described in the documentation you propose to remove 😅

PR #697, opened 2025-02-06
🐛(mailbox) fix status of some mailboxes in production

sdemagny:
All mailboxes imported from "mail.numerique.gouv.fr" and "data.gouv.fr" must be in an enabled state and not pending.
Use a data migration to fix it.

### DIFF
--- a/src/backend/mailbox_manager/migrations/0018_fix_mailboxes_status.py
+++ b/src/backend/mailbox_manager/migrations/0018_fix_mailboxes_status.py
@@ -1,0 +1,23 @@
+from django.db import migrations
+
+from mailbox_manager import enums
+
+
+def change_some_mailboxes_status_to_enabled(apps, schema_editor):
+    Mailbox = apps.get_model('mailbox_manager', 'Mailbox')
+    Mailbox.objects.filter(
+        status=enums.MailboxStatusChoices.PENDING,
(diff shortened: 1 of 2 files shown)


src/backend/mailbox_manager/migrations/0018_fix_mailboxes_status.py
```diff
@@ -0,0 +1,23 @@
+from django.db import migrations
+
+from mailbox_manager import enums
+
+
+def change_some_mailboxes_status_to_enabled(apps, schema_editor):
+    Mailbox = apps.get_model('mailbox_manager', 'Mailbox')
+    Mailbox.objects.filter(
+        status=enums.MailboxStatusChoices.PENDING,
+        domain__name__in=["mail.numerique.gouv.fr", "data.gouv.fr"]
```
mjeammet, line 10: I wonder if domain names in clear poses a security issue or not at all

mjeammet (approved):
Thanks !

sdemagny:
On s'est mal compris. Ça part en script.

src/backend/mailbox_manager/migrations/0018_fix_mailboxes_status.py
sdemagny: No but it's too specific. I will send a script to Jacques ;)

PR #708, opened 2025-02-10
✨(api) allow invitations for domain management 

mjeammet:
## Purpose

add an endpoint to llow to invite someone to join people
and manage a domain

## Proposal

Description...

- [x] create /mail-domains/<slug>/invitations endpoint
- [x] automatically add domain to invitation <3
- [x] tests on invitation creation
- [x] tests on invitation listing
- [x] tests on invitation retrieval
- [x] when someone uses an invitation to join People, their invitations are automatically transformed into domain accesses


src/backend/mailbox_manager/api/client/serializers.py
```diff
@@ -246,3 +246,41 @@ class Meta:
...
+        user = getattr(request, "user", None)
+
+        try:
+            domain_slug = self.context["domain_slug"]
+        except KeyError as exc:
+            raise exceptions.ValidationError(
+                "You must set a domain slug in kwargs to create a new domain management invitation."
+            ) from exc
+
+        if not models.MailDomainAccess.objects.filter(
```
sdemagny, line 272: domain.get_abilities(user) -> manage_access = True

qbey (approved):

PR #711, opened 2025-02-11
🚨(pytest) Fix or suppress warnings during backend tests

Morendil:
## Purpose

Avoid unnecessary noise from testing processes: when testing, a green bar should actually be green, not orange.

## Proposal

Fix or suppress warnings from `make test`.


src/backend/core/api/client/viewsets.py
```diff
@@ -4,6 +4,7 @@
 import operator
 from functools import reduce
 
+import django.utils.timezone
```
qbey, line 7: Nit: We usually use `from django.utils import timezone`

src/backend/mailbox_manager/admin.py
```diff
@@ -79,10 +82,16 @@ def fetch_domain_status_from_dimail(modeladmin, request, queryset):  # pylint: d
...
         ]
-        messages.success(request, format_html("<br> ".join(map(str, msg_success))))
+        messages.success(
+            request,
+            format_html_join(mark_safe("<br> "), "{}", ([str(m)] for m in msg_success)),
+        )
     if msg_error:
         msg_error.insert(0, _("Check domain failed for:"))
-        messages.error(request, format_html("<br> ".join(map(str, msg_error))))
+        messages.success(
```
qbey, line 91:
```suggestion
        messages.error(
```

qbey (commented):
Thanks for the cleanup

src/backend/mailbox_manager/admin.py
Morendil: Ooops ! Well spotted.

mjeammet (approved):
Ouh ça soulage !

PR #719, opened 2025-02-12
:loud_sound:(prod) move logging config up to Base configuration class

Morendil:
## Purpose

Make it possible to set logging configuration on a per-deployment basis in production, so we can actually see our application logs in our production environment.

## Proposal

Currently, LOGGING is defined only on the Development class. This means that the Staging, PreProduction and Production classes, which are used in our prod environments, do not define any logger options. As a result, our production environments use the default Python logging configuration: a log level of WARNING, and no timestamping of logs, among other things.

This PR moves up the LOGGING definition, to the Base class, this way Production and subclasses will inherit it. This will make INFO the default log level for the root and app loggers. This definition includes an environment variable mapping, so we can still configure a different log level for those if necessary.


qbey (approved):
Nice fix :)

PR #729, opened 2025-02-14
🔧(swagger) activate swagger in staging

mjeammet:
## Purpose

activate swagger in staging for devs from other teams to build interop more easily
Activate ONLY on staging because the memory leaks issue is **not** fixed. See #673 

## Proposal

(could it simply be a USE_SWAGGER = True ? 🤔 suspicious)


Morendil:
Apparently it can ! It could be useful to make it a value controlled by an environment variable, this way the choice could then be made in the deployment repo for any environment.

Morendil (approved):

PR #757, opened 2025-03-04
🐛(mailbox) fix migration to fill dn_email field

sdemagny:
Fix AttributeError
'Mailbox' object has no attribute 'get_email'


qbey (approved):
thanks !

PR #762, opened 2025-03-05
🔧(dns) make target zone for communes domains configurable

Morendil:
## Purpose

Ease management of subdomains created for testing purposes.

## Proposal

Currently domains for communes created during testing share the same destination (collectivite.fr) as the ones created during production.

Add a configuration setting tied to an env var, so we can have a separate zone for staging/preprod.

This way it's easier to tell which subdomains are for testing and can be deleted without user impact.

Default value is None to keep code generic.

### DIFF
--- a/src/backend/people/settings.py
+++ b/src/backend/people/settings.py
@@ -545,6 +545,11 @@ class Base(Configuration):
         environ_name="MAIL_CHECK_DOMAIN_INTERVAL",
         environ_prefix=None,
     )
+    DNS_PROVISIONING_TARGET_ZONE = values.Value(
+        default=None,
+        environ_name="DNS_PROVISIONING_TARGET_ZONE",
+        environ_prefix=None,
+    )
     DNS_PROVISIONING_API_URL = values.Value(
         default="https://api.scaleway.com",
         environ_name="DNS_PROVISIONING_API_URL",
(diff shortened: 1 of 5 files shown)


src/backend/people/settings.py
```diff
@@ -512,6 +512,11 @@ class Base(Configuration):
         environ_name="MAIL_PROVISIONING_API_CREDENTIALS",
         environ_prefix=None,
     )
+    DNS_PROVISIONING_TARGET_ZONE = values.Value(
+        default="collectivite.fr",
```
qbey, line 516: Since its in common settings, please don't use this default :)

Morendil: @qbey What's a suitable alternative ? Note that the variable is only used by the plugin anyway, so unlike common code which would impact all users, this particular setting has no effect on anyone who does not use the plugin.

Morendil:
@qbey Take 2, does this work for you as amended ? My eyes are bleeding a little bit from having conditional code depend on a global variable, but in the spirit of QR/QC I would prefer to merge an immediate improvement to our operational tidiness, and await a more comprehensive plugins approach later.

Morendil:
@qbey Take 3… does that work ?

qbey (approved):

PR #767, opened 2025-03-06
✨(api) define dimail timeout as a setting

sdemagny:
In use a timeout of 30 seconds seems more appropriate

### DIFF
--- a/src/backend/mailbox_manager/utils/dimail.py
+++ b/src/backend/mailbox_manager/utils/dimail.py
@@ -41,6 +41,7 @@ class DimailAPIClient:
 
     API_URL = settings.MAIL_PROVISIONING_API_URL
     API_CREDENTIALS = settings.MAIL_PROVISIONING_API_CREDENTIALS
+    API_TIMEOUT = settings.MAIL_PROVISIONING_API_TIMEOUT
 
     def get_headers(self, user_sub=None):
         """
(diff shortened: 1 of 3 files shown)


qbey (commented):
This should be a setting :)

sdemagny:
> This should be a setting :)

Yes! I totally agree ;)

Morendil (approved):

Morendil:
Agree with @qbey and it raises lots of questions - this can potentially be a setting for each call to an external API. From a QR/QC standpoint it makes sense to me to increase timeouts as a quick fix if we're seeing issues in prod, and start thinking about a more comprehensive solution.

src/backend/mailbox_manager/utils/dimail.py
```diff
@@ -35,13 +35,14 @@
...
 
+API_URL = settings.MAIL_PROVISIONING_API_URL
+API_CREDENTIALS = settings.MAIL_PROVISIONING_API_CREDENTIALS
+API_TIMEOUT = settings.MAIL_PROVISIONING_API_TIMEOUT
+
 
 class DimailAPIClient:
     """A dimail-API client."""
 
-    API_URL = settings.MAIL_PROVISIONING_API_URL
```
qbey, line 42: I prefer when settings comes from the class, it makes changes for test or script easier ^^

PR #776, opened 2025-03-07
🔒️(organization) the first user is not admin

### LINKED ISSUE #775: Remove "first user is organization admin" feature
The first user of an organization should not be given the admin permission.

qbey:
## Purpose

The first user of a organization is probably not an admin. This was implemented for first tests but for now it's more a security issue than something helpful.

FIXES #775

## Proposal

- [x] Do not associate organization admin access when creating an Organization at user login.


sdemagny (approved):

Morendil (changes requested):
We are currently relying on this behaviour in the communes plugin, so I would prefer delaying this change until the reworked onboarding process is ready.

PR #827, opened 2025-03-17
:children_crossing:(communes) improve user experience for auto-provisioned domains

Morendil:
## Purpose

In actual user testing we found that making end users responsible for technical operations on the domains that we ourselves create (to the exact specifications of Dimail) was confusing and blocked them from progressing in the onboarding process.

This PR aims to fix this while preserving the property that domains created by Régie are checked by Dimail. This needs to be a scheduled task because a synchronous check leaves too little time for the newly created zone to be advertised by the DNS hoster.

## Proposal

Create commune domaines in the Active state with a short term scheduled check.


jblagadec:
à mettre en stand-by pour l'instant ; on prend le parti de mettre toutes les spécificités de l'on-boarding ST côté site

PR #832, opened 2025-03-19
✨(oidc) add simple introspection backend

qbey:
## Purpose

This provides a configurable OIDC introspection backend to be able to
call introspection endpoints which returns JSON data instead of an
encrypted JWT.

Two backends are currently defined:

 - ResourceServerBackend` which expect a JSON response
 - JWTResourceServerBackend which implements RFC 9701 and expects
   JWE reponse.

There might be other cases (eg: ResourceServerBackend with JWT, JWS or
JWE, etc. but for now we don't use it, so we follow YAGNI).

This also allow to configure the claim to determine the "audience":

 - client_id: for our Keycloak implementation
 - aud: used by ProConnect

## Proposal

- [x] implement the simple backend
- [x] update tests


qbey:
Note: this allow local testing with keycloak and should preserve production use with proconnect.

PR #833, opened 2025-03-19
♻️(dimail) refacto setup_dimail_db to call dimail client

mjeammet:
## Purpose

Management command "setup_dimail_db" called dimail directly, thus creating duplicated code. It now calls "create_domain" and "create_allow" methods from DimailAPIClient (create_user is left unchanged to create special users such as dimail admin or people)

## Proposal

Description...

- [x] use create_allow and create_domain from DimailAPIClient
- [x] fixed inconsistencies in tests

I keep using setup_dimail_db's "create_user" because it allows for admin=True for admin user and for special perms for people user. Client's method on the contrary, does not, which makes sense because we want all the dimail users created from people to be basic (== admin=False + perms = []). 

Rather than modifying client's method, I sticked to the one in setup_dimail_db but we can discuss that.

### DIFF
--- a/src/backend/mailbox_manager/utils/dimail.py
+++ b/src/backend/mailbox_manager/utils/dimail.py
@@ -86,7 +86,7 @@ class DimailAPIClient:
         payload = {
             "name": domain_name,
             "context_name": domain_name,  # for now, we put each domain on its own context
-            "features": ["webmail", "mailbox"],
+            "features": ["webmail", "mailbox", "alias"],
             "delivery": "virtual",
         }
         try:
@@ -115,7 +115,7 @@ class DimailAPIClient:
 
         return self.raise_exception_for_unexpected_response(response)
 
-    def create_mailbox(self, mailbox, user_sub=None):
+    def create_mailbox(self, mailbox, request_user=None):
         """Send a CREATE mailbox request to mail provisioning API."""
 
         payload = {
@@ -126,7 +126,7 @@ class DimailAPIClient:
             # displayName value has to be unique
             "displayName": f"{mailbox.first_name} {mailbox.last_name}",
         }
-        headers = self.get_headers(user_sub)
+        headers = self.get_headers(request_user)
 
         try:
             response = session.post(
@@ -148,7 +148,7 @@ class DimailAPIClient:
             logger.info(
                 "Mailbox successfully created on domain %s by user %s",
                 str(mailbox.domain),
-                user_sub,
+                request_user,
             )
             return response
 
--- a/Makefile
+++ b/Makefile
@@ -296,7 +296,7 @@ recreate-dimail-container:
 
 dimail-setup-db:
 	@echo "$(BOLD)Populating database of local dimail API container$(RESET)"
-	@$(MANAGE) setup_dimail_db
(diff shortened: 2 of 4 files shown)


src/backend/mailbox_manager/utils/dimail.py
```diff
@@ -115,7 +115,7 @@ def create_domain(self, domain_name, request_user):
 
         return self.raise_exception_for_unexpected_response(response)
 
-    def create_mailbox(self, mailbox, user_sub=None):
+    def create_mailbox(self, mailbox, request_user=None):
```
Morendil, line 118: LGTM overall, my only nitpick would be the name of this parameter. 

We are getting into a state where the same value is inconsistently called "request_user" in some contexts (callers like this method) and "user_sub" in others (formal parameter of get_headers, where it is associated with a "username" key, adding to the confusion).

The meaning of these qualifiers of "user" is not clear. For instance if we use "request" then a reader new to the code base will look at the surrounding code and probably assume it is related to the "requests" library - which I'm not sure is a correct assumption. The "sub" is something I assume has to do with OIDC but it's not made clear.

For the purpose of this PR I suggest using `user_sub` to be more like the surrounding code (in QR mode) and I would suggest a comprehensive review of this module (in QC mode).

Morendil (approved):

src/backend/mailbox_manager/utils/dimail.py
mjeammet: If those two variables are indeed subs, but I use them with two different meanings :
- `user_sub` is a payload data sent to dimail to create an object (a user or a permission between user and domain). We chose to use the ProConnect sub because it's supposed to be immutable, where name or email are not.
- `request_user` is the issuer of the request. Some dimail actions require permission on dimail's side and we verify that the aforementioned user_sub+domain object exists. This is more security-related.

I wanted to differentiate those two to be able to log that 
`"request user" created a permission for "another user" on "thatdomain.com"` but maybe it's not worth it if it's not intelligible.

mjeammet: also, what are QR and QC mode ?

Morendil: I'm using these terms [from the Lean lingo](https://www.amelioration.fr/blog/qrqc-def-comprendre-la-methode-rapide-de-resolution-des-problemes-de-qualite), QR for Quick Response and QC for Quality Control. Not as formally as would be the case if practicing Lean explicitly but as a way to remind myself (and others) that a quick fix is often not enough and after detecting problem it's good to take a step back and look at the bigger picture from the perspective of preventing similar problems in the future.

In this context QC would be a team effort to improve the Dimail client.

I think I understand the distinction you're making but I worry that the code does not make it clear, and the next person to make changes might get confused.

Morendil: Thanks for taking the in-person time to review together, I have a better understanding. We're using the requesting user's identity both as a security or traceability feature (log on whose behalf the action is executed). Also, because Dimail makes it mandatory to use a token for certain operations.

We've identified a further refactoring that could be useful: tie the requesting user's info to the Dimail client's constructor. But that can be for a later PR.

PR #836, opened 2025-03-20
🧑‍💻(admin) provider selftest views

qbey:
## Purpose

This allows to define views available in /admin/selftest/ like an email sender testing view for instance.

I'm still not sure about this...

## Proposal

- [x] provide base selftest views
- [x] provider an index view for selftests
- [x] add an example selftest : email
- [ ] add tests


qbey:
Must move this to `django-lasuite`

PR #846, opened 2025-03-27
🔒️(users) restrict listable users to same organization

mjeammet:
## Purpose

This is a quick fix to a security issue. Previously, any user could list all users. Now /users/ endpoint only lists users from same organization.

We might improve this filter later but a quick fix would be good.

## Proposal

Description...

- [x] filter users to list only users from same organization
- [x] fix related tests

### DIFF
--- a/src/backend/core/api/client/viewsets.py
+++ b/src/backend/core/api/client/viewsets.py
@@ -263,9 +263,10 @@ class UserViewSet(
         queryset = self.queryset
 
         if self.action == "list":
-            # Exclude inactive contacts
+            # Filter active users
+            # and users from same organization
             queryset = queryset.filter(
-                is_active=True,
+                is_active=True, organization_id=self.request.user.organization_id
             )
 
             # Exclude all users already in the given team
--- a/src/backend/core/tests/users/test_api_users_list.py
+++ b/src/backend/core/tests/users/test_api_users_list.py
@@ -31,26 +31,29 @@ def test_api_users_list_anonymous():
 
 def test_api_users_list_authenticated():
     """
-    Authenticated users should be able to list all users.
+    Authenticated users should be able to list users from their organization.
     """
-    user = factories.UserFactory()
+    organization = factories.OrganizationFactory(with_registration_id=True)
+    user = factories.UserFactory(organization=organization)
 
     client = APIClient()
     client.force_login(user)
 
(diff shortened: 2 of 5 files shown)


src/backend/core/api/client/viewsets.py
```diff
@@ -267,6 +267,8 @@ def get_queryset(self):
             queryset = queryset.filter(
                 is_active=True,
             )
+            if organization := self.request.user.organization:
```
qbey, line 270: You can remove a query by using the 'user.organization_id' then filter with pk.
This way django won't make a query to fetch all organization's data.

src/backend/core/tests/users/test_api_users_list.py
```diff
@@ -74,7 +76,7 @@ def test_api_users_list_authenticated_response_content(
         language="fr-fr",
     )
 
-    with django_assert_num_queries(3):  # get User, Count, Users
+    with django_assert_num_queries(4):  # get User, Count, Users
```
qbey, line 79: Should be back to 3 after previous comment.

qbey (commented):
Thanks!

src/backend/core/api/client/viewsets.py
mjeammet: works like a charm but, to lower the number of queries I have to add this filter to the active user filter.
Can I safely remove the "if" on line 270 ? Do all users have one an only one organization ? This is not the case in prod but we can fix these discrepancies

mjeammet:
e2e test were crashing because users were not visible anymore (because not in same orga). I've tried adding all e2e users to the same orga but it doesn't seem to fix anything. 

I'm having a hard time running e2e tests on my computer, can someone help me close this PR this week, pliz ? @Morendil @sdemagny @qbey

src/backend/core/tests/users/test_api_users_list.py
```diff
@@ -33,17 +33,19 @@ def test_api_users_list_authenticated():
     """
     Authenticated users should be able to list all users.
     """
-    user = factories.UserFactory()
+    organization = factories.OrganizationFactory(with_registration_id=True)
```
Morendil, line 36: Remember to update the doc comments…

elvoisin (approved):

PR #848, opened 2025-03-28
🚧(plugin) add organization one time token

qbey:
## Purpose

Description...

## Proposal

Description...

- [] item 1...
- [] item 2...


qbey:
Won't do

PR #854, opened 2025-03-31
🧑‍💻(pre-commit) add pre-commit hooks

qbey:
## Purpose

Enforce back and front linter to run before commit.

This could be improved:
 - we could look at https://pre-commit.com/

## Proposal

- [x] add command to install hooks
- [x] add back and front linter run in pre-commit hook

I did not use https://pre-commit.com/ to keep it simple:
 - using pre-commit.com would required a dedicated docker image (python required)
 - but this would also enable the use of a lot of existing hooks

Possible improvements:
 - better linting error display when used as a hook.
 - I don't like we use docker instance to run the hooks, but it allows to run them without local specific environment.

### DIFF
--- a/scripts/git-hooks/pre-commit
+++ b/scripts/git-hooks/pre-commit
@@ -1,0 +1,79 @@
+#!/bin/bash
+
+# Variable to check if we are in script mode of the pre-commit hook
+# This is used to determine if we should run the script in a specific mode
+# or if we are running it as a pre-commit hook
+# Get value from arguments or default to 0
+SCRIPT_MODE=${1:-0}
+
+if [ ${SCRIPT_MODE} -eq 1 ]; then
+  echo "Script mode is enabled. Running in pre-commit hook mode."
+fi
+
+FAILED=0
+
+# Colors for output
+if [ ${SCRIPT_MODE} -eq 1 ]; then
+  # Only set colors if output is a terminal
+  BOLD="\033[1m"
+  GREEN="\033[1;32m"
+  RED="\033[1;31m"
+  RESET="\033[0m"
+
+  GIT_CACHED=""
+else
+  # Fallback to no colors if not a terminal
+  BOLD=""
+  GREEN=""
+  RED=""
+  RESET=""
+
+  GIT_CACHED="--cached"
+fi
+
+echo "${BOLD}Running pre-commit hooks...${RESET}"
+
+# Check if there are staged changes in Python files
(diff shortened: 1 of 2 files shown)


scripts/git-hooks/pre-commit
```diff
@@ -0,0 +1,52 @@
...
+  fi
+else
+    echo "${BOLD}No Python files staged for commit. Skipping backend linters...${RESET}"
+fi
+
+# Check if there are staged changes in frontend files
+FRONT_STAGED_FILES=$(git diff --name-only --diff-filter=ACM | grep -E "src/frontend/.*\.(js|jsx|ts|tsx)$")
+if [ -n "$FRONT_STAGED_FILES" ]; then
+  echo "${BOLD}Running frontend linters...${RESET}"
+  make lint-front
```
Morendil, line 34: Unlike the backend linter, the frontend linter doesn't write its improvements, this should be the default for frontend linting as well.

qbey:
@mjeammet This is not perfect, but is could be a first step for the hooks. You may try this and tell me what do you think.

scripts/git-hooks/pre-commit
qbey: This is out of scope for this PR, but feel free to add commits on this if you see things which can be improved of course

scripts/git-hooks/pre-commit
```diff
@@ -0,0 +1,75 @@
...
+
+# Check if there are staged changes in Python files
+PYTHON_STAGED_FILES=$(git diff ${GIT_CACHED} --name-only --diff-filter=ACM | grep -E "\.py$")
+if [ -n "$PYTHON_STAGED_FILES" ]; then
+  echo "${BOLD}Running backend linters...${RESET}"
+  make lint
+  BACKEND_RESULT=$?
+  if [ $BACKEND_RESULT -ne 0 ]; then
+    echo "${RED}Backend linting failed!${RESET}"
+    echo "${RED}Please fix the issues and try committing again.${RESET}"
```
mjeammet, line 44: Looks like the RED isn't taken into account, even tho I don't see why not, because GREEN works
image
****

mjeammet: both !

mjeammet:
It sounded like a good idea during the retro but I'm wondering now. Especially when 
image

qbey:
I will drop this, I don't like the fact theses hooks are too heavy and for now, running make commands manually is not that complex.

PR #855, opened 2025-03-31
List SIRET endpoint | account service 

sdemagny:
Reste à faire :
- [x] check permission scope
- [x] fix INSTALLED_PLUGINS load


src/backend/core/authentication/backends.py
```diff
@@ -245,3 +248,45 @@ def update_user_if_needed(self, user, claims):
 
         if updated_claims:
             self.UserModel.objects.filter(sub=user.sub).update(**updated_claims)
+
+
+class AccountServiceAuthentication(BaseAuthentication):
```
qbey, line 253: Why don't you inherit from TokenAuthentication to reduce code?

src/backend/core/models.py
```diff
@@ -1091,3 +1092,33 @@ def get_abilities(self, user):
...
+
+class AccountService(BaseModel):
+    """Account service model."""
+
+    name = models.CharField(_("name"), max_length=255)
+    api_key = models.CharField(
+        _("api key"),
+        max_length=255,
+        default=secrets.token_urlsafe(32),
+        editable=False,
```
qbey, line 1105: Why not?

src/backend/people/settings.py
```diff
@@ -526,6 +526,14 @@ class Base(Configuration):
...
     )
 
+    ACCOUNT_SERVICE_SCOPES = values.ListValue(
+        default=[
+            ("la-suite-list-organizations-siret", "List organizations with SIRET"),
+        ],
+        environ_name="ACCOUNT_SERVICE_SCOPES",
+        environ_prefix=None,
+    )
+
```
qbey, line 536: Only used in the model char field choices:
 - This should be dynamic in the admin maybe
 - The default should not be our custom plugin

This can be kept like this for simplicity sake but please remove the default.

src/backend/plugins/la_suite/api/viewsets.py
```diff
@@ -0,0 +1,71 @@
...
+    ViewSet to list all SIRET of active communes.
+
+    * Requires API key authentication.
+    * Only services with the correct scope are able to access this view.
+    """
+
+    authentication_classes = [AccountServiceAuthentication]
+    permission_classes = [ScopeAPIPermission]
+    pagination_class = Pagination
+    scope = "la-suite-list-organizations-siret"
```
qbey, line 42: Should be a constant?

src/backend/plugins/la_suite/api/viewsets.py
```diff
@@ -0,0 +1,71 @@
...
+        registration_ids_lists = self.get_queryset().values_list(
+            "registration_id_list", flat=True
+        )
+        registration_ids = [
+            registration_id
+            for registration_ids_list in registration_ids_lists
+            for registration_id in registration_ids_list
+        ]
+
+        page = self.paginate_queryset(registration_ids)
```
qbey, line 67: The pagination purpose is also to reduce the database stress, here you load every siret every time and only return a subset. 

Maybe you should return `registration_id_list` directly and let the consumer do the job?

src/backend/plugins/la_suite/urls.py
```diff
@@ -0,0 +1,14 @@
+"""API URL Configuration for La Suite plugin"""
+
+from rest_framework.routers import DefaultRouter
+
+from .api.viewsets import ActiveOrganizationsSiret
+
+router = DefaultRouter()
+router.register(
+    r"la-suite/v1.0/siret",
```
qbey, line 9:
```suggestion
    "la-suite/v1.0/siret",
```

src/backend/plugins/la_suite/urls.py
```diff
@@ -0,0 +1,14 @@
...
+    ActiveOrganizationsSiret,
+    basename="active-organizations-siret",
```
qbey, line 11:
```suggestion
    basename="active-organization-sirets",
```
?

env.d/development/common.dist
```diff
@@ -65,3 +65,4 @@ IH5em4M/pHIcsqCi1qggBMbdvzHBUtC3R4sK0CpEFHlN+Y59aGazidcN2FPupNJv
 MbyqKyC6DAzv4jEEhHaN7oY=
 -----END PRIVATE KEY-----
 "
+INSTALLED_PLUGINS=plugins.la_suite.apps.LaSuitePluginConfig
```
qbey, line 68:
```suggestion

```

src/backend/core/models.py
sdemagny: Because it's an automatically generated token. It makes sense to me. But whatever.

src/backend/plugins/la_suite/urls.py
sdemagny: j'avais vu que c'était invariant, j'ai évité de mettre du sirets. Mais je n'arrive pas à retrouver l'info :D

qbey (approved):

PR #860, opened 2025-04-02
🐛(plugin) allow simple application name

qbey:
## Purpose

This allows to use the application name, instead of the full path to the application configuration in the INSTALLED_PLUGINS setting.

## Proposal

Use a mix in for the application configuration.


sdemagny (approved):
Thanks! 🙏

PR #864, opened 2025-04-03
🐛(stats) rename stat for clarity

mjeammet:
## Purpose

Public statistics on domains was modified to count only enabled domains. Modify stat name to reflect change.

## Proposal

Description...

- [x] rename stat on both enpoind and tests

### DIFF
--- a/src/backend/core/tests/test_api_stats.py
+++ b/src/backend/core/tests/test_api_stats.py
@@ -30,7 +30,7 @@ def test_api_stats__anonymous(django_assert_num_queries):
     assert response.json() == {
         "total_users": 0,
         "mau": 0,
-        "domains": 5,
+        "active_domains": 5,
         "mailboxes": 0,
         "teams": 3,
     }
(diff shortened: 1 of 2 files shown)


src/backend/core/tests/test_api_stats.py
```diff
@@ -30,7 +30,7 @@ def test_api_stats__anonymous(django_assert_num_queries):
     assert response.json() == {
         "total_users": 0,
         "mau": 0,
-        "domains": 5,
+        "active domains": 5,
```
Morendil, line 33: Consistency objection: the resulting code spells "total users" with an underscore and "active domains" with a space.

Spaces are more legible, and can make parsing harder from some oddball languages. Underscores are more portable but less aesthetic.

mjeammet: Damn you're right 🤦🏽 seems like I can't focus these days. I'll go for underscores as aesthetic doesn't matter much here.

qbey (approved):

PR #879, opened 2025-04-10
✨(mailbox) remove secondary email as required field

sdemagny:
The secondary email address is no longer required for all creation processes and we should not force the user to provide and store an insecure email address.


sylvinus:
Great!

sylvinus (approved):

PR #887, opened 2025-04-14
🔒️(passwords) add HIBP password validator to catch more bad passwords

Morendil:
## Purpose

Add HIBP password validator to catch more bad passwords

## Proposal

We set the "fail on error" to False for initial experiments.


Morendil:
Not quite enough bandwidth at the moment to bring this to completion, will pick it up later when we resume work on security actions for homologation.

PR #889, opened 2025-04-23
✨(mailbox) synchronize password of newly created mailbox with Dimail's

Morendil:
## Purpose

When using La Régie as Identity Provider this allows signing in.

## Proposal

As a temporary measure, set the mailbox password to be the one created by Dimail. (Note that this raises further questions, such as what means we can provide for the user to set a different password.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


qbey (approved):

qbey:
@mjeammet you may merge when you want :)

PR #895, opened 2025-04-30
✨(resource-server) add SCIM `/Me` endpoint

qbey:
## Purpose

This provides a "self-care" SCIM endpoint, authenticated with OIDC token introspection. This endpoint will be use by services to fetch the user's team list.

We chose to use the SCIM format (even if this is not a SCIM context) to make it easier to understand/maintain/plug.

## Proposal

Add a new resource server authenticated SCIM formatted endpoint: `/Me`

- [x] add `/Me` endpoint
- [x] add tests for response format content

### DIFF
--- a/src/backend/people/resource_server_urls.py
+++ b/src/backend/people/resource_server_urls.py
@@ -3,18 +3,23 @@
 from django.urls import include, path
 
 from lasuite.oidc_resource_server.urls import urlpatterns as resource_server_urls
-from rest_framework.routers import DefaultRouter
+from rest_framework.routers import SimpleRouter
 
 from core.api.resource_server import viewsets
+from core.api.resource_server.scim import viewsets as scim_viewsets
 
 # - Main endpoints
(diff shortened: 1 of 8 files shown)


src/backend/people/resource_server_urls.py
```diff
@@ -3,18 +3,23 @@
...
 # - Main endpoints
 # Contacts will be added later
 # Users will be added later
-router = DefaultRouter()
+router = SimpleRouter()
 router.register("teams", viewsets.TeamViewSet, basename="teams")
 
+# - SCIM endpoints
+scim_router = SimpleRouter()
+scim_router.register("Me", scim_viewsets.MeViewSet, basename="scim-me")
```
lunika, line 19: Why not `/me` ?

qbey: Because... SCIM : https://scim.dev/playground/me

PR #901, opened 2025-05-06
✨(admin) send pending mailboxes from admin

mjeammet:
## Purpose

Adds an admin action to send pending mailboxes on an active domain.
This is for debug purposes when mailbox creation failed in dimail and we end up with pending mailboxes.

## Proposal

Description...

- [x] add admin action
- [x] modify "enable_pending_mailboxes" to return mailboxes for which sending failed
- [x] add tests

### DIFF
--- a/src/backend/mailbox_manager/admin.py
+++ b/src/backend/mailbox_manager/admin.py
@@ -135,6 +135,39 @@ def fetch_domain_expected_config_from_dimail(modeladmin, request, queryset):  # 
         )
 
 
+@admin.action(description=_("Send pending mailboxes to dimail"))
+def send_pending_mailboxes(modeladmin, request, queryset):  # pylint: disable=unused-argument
+    """Send pending mailboxes"""
+    client = DimailAPIClient()
+
+    excluded_domains = []
+    for domain in queryset:
+        # do not check disabled domains
+        if domain.status != enums.MailDomainStatusChoices.ENABLED:
+            excluded_domains.append(domain.name)
+            continue
+
+        results = client.send_pending_mailboxes(domain)
+        if failed_mailboxes := results["failed_mailboxes"]:
+            messages.error(
+                request,
+                _("Failed to send the following mailboxes : %(mailboxes)s.")
+                % {"mailboxes": ", ".join(failed_mailboxes)},
+            )
+        else:
+            messages.success(
+                request,
+                _("Pending mailboxes successfully sent for %(domain)s.")
+                % {"domain": domain.name},
+            )
+    if excluded_domains:
+        messages.warning(
+            request,
+            _("Domains disabled are excluded from : %(domains)s")
+            % {"domains": ", ".join(excluded_domains)},
+        )
+
+
 class UserMailDomainAccessInline(admin.TabularInline):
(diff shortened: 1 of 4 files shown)


src/backend/mailbox_manager/tests/test_admin_actions.py
```diff
@@ -204,3 +204,83 @@ def test_fetch_domain_expected_config__should_not_fetch_for_disabled_domain(clie
     assert "Domains disabled are excluded from fetch" in response.content.decode(
         "utf-8"
     )
+
+
+@responses.activate
+@pytest.mark.django_db
+def test_send_pending_mailboxes(client):
```
qbey, line 211: You may add an already enabled domain to check the message about already activated

src/backend/mailbox_manager/admin.py
```diff
@@ -135,6 +135,39 @@ def fetch_domain_expected_config_from_dimail(modeladmin, request, queryset):  #
...
+
+    excluded_domains = []
+    for domain in queryset:
+        # do not check disabled domains
+        if domain.status != enums.MailDomainStatusChoices.ENABLED:
+            excluded_domains.append(domain.name)
+            continue
+
+        response = client.enable_pending_mailboxes(domain)
+        if response:
```
qbey, line 151: I think this is always "True" because response is `{"failed_mailboxes": failed_mailboxes}`

I suggest to name "response" => "results" and therefore:
```suggestion
        if failed_mailboxes := results["failed_mailboxes"]:
```

src/backend/mailbox_manager/tests/test_admin_actions.py
mjeammet: I'm not sure which "already activated" message are you referring to. 

As this tests simulate a situation where mailboxes that fell out of sync on an enabled domain, I already use an enabled domain (+ pending mailboxes).

Did you want me to add a "nothing to sync" message when the domain is enabled + 0 pending mailboxes ?

src/backend/mailbox_manager/admin.py
```diff
@@ -135,6 +135,39 @@ def fetch_domain_expected_config_from_dimail(modeladmin, request, queryset):  #
...
+                request,
+                _("Failed to send the following mailboxes : %(mailboxes)s.")
+                % {"mailboxes": ", ".join(response["failed_mailboxes"])},
+            )
+        else:
+            messages.success(
+                request,
+                _("Pending mailboxes succesfully sent for %(domain)s.")
+                % {"domain": domain.name},
+            )
```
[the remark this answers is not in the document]
mjeammet: This generated one line per domain. 
Capture d’écran du 2025-05-15 15-17-46
Would you rather have all these info on a single message ?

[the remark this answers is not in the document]
qbey: Hum, no you are right, I guess one line per domain is ok :)

src/backend/mailbox_manager/tests/test_admin_actions.py
qbey: Ignore my comment, I made a confusion between mail and domain I guess ^^

qbey (approved):

PR #909, opened 2025-05-16
🎨(front) fix mail domain list display

qbey:
## Purpose

The first implementation was using `div` instead of the proper components.

## Proposal

Replace `div` with `Box` and `h2` with `Text`


qbey:
Merged

PR #937, opened 2025-06-27
🔧(back) remove usage of deprecated db engine

lunika:
## Purpose

The db engine postgresql_psycopg2 does not exists anymore in django but for BC compat it is possible to use it in the configuration and it is replace by postgresql at runtime. We changed this settings to use the good one.

## Proposal

- [x] 🔧(back) remove usage of deprecated db engine


qbey (approved):
thx

PR #949, opened 2025-07-02
🧑‍💻(demo) add mailboxes to demo

mjeammet:
## Purpose

add mailboxes to demo domains.

## Proposal

Description...

- [x] add a few mailboxes to every domains
- [x] add more than 20 mailboxes to a specific many-boxed domain (can act as a local test for pagination)
- [x] prepare invitations for John Doe :)

### DIFF
--- a/src/backend/demo/management/commands/create_demo.py
+++ b/src/backend/demo/management/commands/create_demo.py
@@ -181,7 +181,7 @@ def create_oidc_people_idp_client_user():
     )
 
 
-def create_demo(stdout):  # pylint: disable=too-many-locals
+def create_demo(stdout):  # pylint: disable=too-many-branches too-many-statements too-many-locals # noqa: PLR0912, PLR0915
     """
     Create a database with demo data for developers to work in a realistic environment.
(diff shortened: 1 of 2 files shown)


src/backend/demo/management/commands/create_demo.py
```diff
@@ -303,6 +303,29 @@ def create_demo(stdout):  # pylint: disable=too-many-locals
...
+                        first_name=first_name,
+                        last_name=last_name,
+                        local_part=local_part,
+                        domain_id=domain_id,
+                        secondary_email=f"{local_part}@example.fr",
+                        status=random.choice(MailboxStatusChoices.values),
+                    )
+                )
+
+        queue.flush()
```
mjeammet, line 327: For reasons, `make demo` fails here because of supposed duplicates in `dn_emails`. By changing randint to (1,1) by example, it's easy to see that we don't have no duplicates (not even on `local_mailbox` parameter). 

@qbey could you please take a look ?

qbey (approved):

Issue #859, opened 2025-04-02
Invitation to manage a domain doesn't succeed if user already exists on another organization


Morendil:
## Bug Report

**Problematic behavior**
When [converting](https://github.com/suitenumerique/people/blob/1eb9dffa4843f694c7aa49371fa713af5fbcaeda/src/backend/mailbox_manager/signals.py#L23) a domain invitation, if the User object already exists, the boolean flag `created` will have the value False, with the result that the user will not be created in Dimail.

When attempting to manage the domain, user will see the dreaded message `Could not retrieve a token, please check 'MAIL_PROVISIONING_API_CREDENTIALS' setting.`

**Expected behavior/code**
The user should be created in Dimail and an allows created; the method should not be a no-op if the user already exists.

**Steps to Reproduce**
This was seen in production by Denis B, who logged in via ProConnect on Apr 1 at 14:36, was invited by Julien B on Apr 2 at 10:24, and was unable to manage the domain after receiving the invitation.

The production logs are consistent with the hypothesis above: we see the line `2025-04-02 08:24:29,343 mailbox_manager.signals INFO Convert domain invitations for user Denis B` but we do not see a line "Invitations converted to domain accesses"… Actually we do not see the confirmation line in the log for any of the other "Convert domain invitations for" lines.

**Environment**
- People version: 1.14.1

**Possible Solution**
Make the body of the `if` unconditional ?

mjeammet:
Thanks @Morendil . I'll take it (but not right now, I have too many open PRs at the moment).

Morendil:
I think the member search box for "Add a new access" restricts the search to users who already have an access on the domain, so Julien would not have seen Denis in the search list.

[...]

PR #953, opened 2025-07-03
🐛(domains) fix invitation instead of access for existing users

mjeammet:
## Purpose

Users can only search other users inside their own organization. This leads to situations where 
- people are trying to share accesses to coworkers from other organization
- the back-end doesn't realize the person already exists
- sends an invite
- invite is never consumed because their account already exists 🙃 
See #859 

## Proposal

Description...

- [x] Fix tests
- [ ] Define if this is a back-end issue or a front-end issue (or both)
- [ ] Fix it
- [ ] manually "convert" potential pending invitations to accesses in prod

### DIFF
--- a/src/backend/mailbox_manager/tests/api/invitations/test_api_domain_invitations_create.py
+++ b/src/backend/mailbox_manager/tests/api/invitations/test_api_domain_invitations_create.py
@@ -11,7 +11,7 @@ from rest_framework.test import APIClient
 
 from core import factories as core_factories
 
-from mailbox_manager import enums, factories
+from mailbox_manager import enums, factories, models
 from mailbox_manager.api.client import serializers
 
 pytestmark = pytest.mark.django_db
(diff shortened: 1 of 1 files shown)


src/backend/mailbox_manager/api/client/viewsets.py
```diff
@@ -392,3 +393,16 @@ def get_queryset(self):
...
+    def perform_create(self, serializer):
+        """Lookup for existing user before inviting."""
+        if email := serializer.validated_data["email"]:
+            existing_user = models.User.objects.filter(email=email)
+            if existing_user.exists():
+                return models.MailDomainAccess.objects.create(
+                    user=existing_user[0],
+                    domain=serializer.validated_data["domain"],
+                    role=serializer.validated_data["role"],
+                )
```
qbey, line 406: To make only one query:
```suggestion
            try:
                existing_user = models.User.objects.get(email=email)
            except models.User.DoesNotExist:
                existing_user = None
            if existing_user:
                return models.MailDomainAccess.objects.create(
                    user=existing_user,
                    domain=serializer.validated_data["domain"],
                    role=serializer.validated_data["role"],
                )
```

mjeammet:
Cannot test interaction with front-end locally as access modal seems broken 🤔 
image

can you check @qbey please ?

PR #959, opened 2025-08-28
Fix imports from dimail

mjeammet:
## Purpose

Fix import from dimail, broken after slight change in response's format.
Upgrade dimail to 0.5.2


qbey (approved):
Yes, I don't know how to configure renovate properly either...

PR #960, opened 2025-08-28
💚(ci) fix changelog CI job

mjeammet:
## Purpose

`git whatchanged` got nominated for removal and now requires an additional option to run, to confirm that we still use it.

@qbey should I send a quick mail or do you want to do it ?


qbey:
@mjeammet , @lunika fixed it here : https://github.com/suitenumerique/django-lasuite/pull/26/commits/ea0da9e4c5a644219ee75570bfd08a421c30d67c you can do like him ;)

qbey (approved):
Thanks :)

PR #966, opened 2025-09-17
🐛(fix) add enabled update your mailbox

elvoisin:
can update your mailbox as viewer


mjeammet:
Works like a charm on my config
image

mjeammet (approved):
the bug seems to already exist on main branch in local (not on any other env) and is not specific to this PR. Bad news for life in general, good news for this PR.

PR #974, opened 2025-10-13
Create and delete aliases

mjeammet:
## Purpose

Create, list and delete aliases for a given domain.

## Proposal

Description...

- [x] models, serializers, viewsets, routes, factories and permissions

### Create
- [x] aliases cannot be created if local part is already used by a mailbox
- [x] mailbox cannot be created if local part is already used by an alias
- [x] dimail method
- [x] if dimail unexpectedly returns 409_CONFLICT, raise a clear error asking for sync

### Read
- [x] list aliases

### Update
No update

### Delete
- [x] delete
- [x] viewers can delete self alias
- [x] dimail method
- [x] if dimail unexpectedly returns 404_NOT_FOUND, return a clear message asking for sync

## Improvements in future PR
- fetch existing aliases from dimail
- add aliases to demo
- modify permissions for users to get a 404 when trying to access a domain they have no access to

### DIFF
--- a/src/backend/mailbox_manager/api/client/viewsets.py
+++ b/src/backend/mailbox_manager/api/client/viewsets.py
@@ -2,7 +2,7 @@
 
 from django.db.models import Q, Subquery
 
-from rest_framework import exceptions, filters, mixins, viewsets
+from rest_framework import exceptions, filters, mixins, status, viewsets
 from rest_framework.decorators import action
 from rest_framework.response import Response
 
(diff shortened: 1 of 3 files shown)


src/backend/mailbox_manager/tests/api/aliases/test_api_aliases_create.py
```diff
@@ -0,0 +1,175 @@
...
+    """User authenticated but not having domain permission should not create aliases."""
+    domain = factories.MailDomainEnabledFactory()
+
+    client = APIClient()
+    client.force_login(core_factories.UserFactory())
+    response = client.post(
+        f"/api/v1.0/mail-domains/{domain.slug}/aliases/",
+        {"local_part": "intrusive", "destination": "intrusive@mail.com"},
+    )
+    assert response.status_code == status.HTTP_403_FORBIDDEN
```
qbey: I think there would be several cases to be tested:

- The user can see the domain, but does not have permission: 403
- The user cannot see the domain: 404

src/backend/mailbox_manager/api/client/serializers.py
```diff
@@ -309,3 +320,41 @@ def validate(self, attrs):
...
+            client.create_alias(alias, self.context["request"].user.sub)
+            return super().create(validated_data)
+
+        return None
+
+    def validate_local_part(self, value):
+        """Validate this local part does not match a mailbox."""
+        if models.Mailbox.objects.filter(
+            local_part=value, domain__slug=self.context["domain_slug"]
+        ):
```
qbey:
```suggestion
        if models.Mailbox.objects.filter(
            local_part=value, domain__slug=self.context["domain_slug"]
        ).exists():
```

src/backend/mailbox_manager/tests/api/aliases/test_api_aliases_create.py
mjeammet: Damn, you're right ! I had totally forgotten about that decision, thanks for reminding me.

PR #1002, opened 2025-12-17
✨(aliases) delete single aliases or all aliases in one call

mjeammet:
## Purpose

two different endpoints to delete aliases : 
- a single alias when specifying a single object
- a query-based way to delete multiple aliases in one action

## Proposal

Description...

- [x] delete single alias with pk
- [x] delete all aliases of a given local_part
- [x] improve (?) the way serializer handles domain

## Later

- allow to delete multiple aliases based on same destination
- if we like the new serializer way, we must apply it to all existing serializers !
- viewers can delete alias pointing to their address

### DIFF
--- a/src/backend/mailbox_manager/api/client/viewsets.py
+++ b/src/backend/mailbox_manager/api/client/viewsets.py
@@ -425,6 +425,9 @@ class AliasViewSet(
 
     DELETE /api/<version>/mail-domains/<domain_slug>/aliases/<alias_pk>/
         Delete targeted alias
+
+    DELETE /api/<version>/mail-domains/<domain_slug>/aliases/?local_part=<local_part>/
+        Delete all aliases of targeted local_part
     """
 
     lookup_field = "pk"
@@ -475,5 +478,28 @@ class AliasViewSet(
                 "Domain out of sync with mailbox provider, please contact our support.",
                 status=status.HTTP_200_OK,
             )
+
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    @action(methods=["DELETE"], detail=False)
+    def delete(self, request, *args, **kwargs):
+        """Bulk delete aliases. Filtering is required and accepted filter is local_part."""
+
+        if "local_part" not in self.request.query_params:
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+
+        local_part = self.request.query_params["local_part"]
+        queryset = self.get_queryset().filter(
+            local_part=local_part
+        )  # Manually call get_queryset to filter by domain and role
+        if not queryset:
+            raise Http404("No Alias matches the given query.")
+
+        # view is bounded to a domain, fetch is from the queryset to spare a dedicated DB request"
+        domain_name = queryset[0].domain.name
+        queryset.delete()
+
+        client = DimailAPIClient()
+        client.delete_multiple_alias(local_part, domain_name)
 
         return Response(status=status.HTTP_204_NO_CONTENT)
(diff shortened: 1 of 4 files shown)


mjeammet:
I'm getting a bit lost in the several ways we could be implementing this feature :
- `/aliases/<local_part>/` and `/aliases/<local_part>/destination/<destination>/` <= the one I wrote now
- `/aliases/<local_part>/<destination>/`, where destination can be `all`. <= a bit better I think but I haven't quite found how to setup my routers to make it work
- `/aliases/?local_part="something"`
- `/aliases/` en précisant dans le body ce qu'on veut supprimer
Kinda like those two because they allow deletion by another criteria (such as destination"formercolleague@mail.com")

Which option do you like best, @qbey ? What do you suggest we do ?

---
After discussion, we settled on providing : 
- `/aliases/<alias_pk>/` to delete a single, known alias
- `/aliases/?query` for funkier uses

src/backend/mailbox_manager/api/client/viewsets.py
```diff
@@ -459,39 +463,40 @@ def get_queryset(self):
...
             )
 
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+    @action(methods=["DELETE"], detail=False)
+    def delete(self, request, *args, **kwargs):
+        """Delete multiple aliases."""
+        client = DimailAPIClient()
+
+        if local_part := self.request.query_params.get("local_part"):
```
qbey: What if this parameter is missing?

Also, I think we could rely on the DRF queryset filtering instead (which would allow to ignore case etc.), but I like (and prefer) the idea of enforcing a filtering (ie you cannot call DELETE without filter to delete everything), but then maybe we should return a 400 if the parameters are not valid. 

My suggestion here is to:
 - update the docstring : "Bulk delete aliases. Filtering is required and accepted filter is `local_part`"
 - If the queryparam `local_part` is missing: return a 400
 - Keep the 404 and the rest like you did

src/backend/mailbox_manager/api/client/viewsets.py
```diff
@@ -459,39 +463,40 @@ def get_queryset(self):
...
+            queryset = self.get_queryset().filter(
+                local_part=local_part
+            )  # Manually call get_queryset to filter by domain and role
+            if not queryset:
+                raise Http404("No Alias matches the given query.")
+
+            domain_name = queryset[0].domain.name
```
qbey: This is ok, but not straightforward so a comment saying "the view is bounded to a domain, fetch is from the queryset to spare a dedicated DB request"

qbey (commented):
With few improvements it can be merged, but:
- I highly recommend to uncouple the alias permissions from the domain one soon, and at least make it explicit in code like `permission_classes = [permissions.DomainResourcePermission]  # warning: rely on domain permissions`
- Add test around the "bulk delete view": anonymous cannot delete, viewer cannot delete, local part not found
- Fix the "reader user can remove themselves from an alias"?

mjeammet:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=suitenumerique_people&pullRequest=1002) **Quality Gate failed**
> 
> Failed conditions  [14.5% Duplication on New Code](https://sonarcloud.io/component_measures?id=suitenumerique_people&pullRequest=1002&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=suitenumerique_people&pullRequest=1002)

SonarQube is freaking out because of tests ... can I deactivate it ? Or should I start fixturing more ?

PR #1005, opened 2026-01-08
refresh domain invitations

mjeammet:
## Purpose

Invitations can now be refreshed, de-expiring them or moving their expiration date further in the future.


src/backend/core/models.py
```diff
@@ -1001,14 +1000,19 @@ def clean(self):
         if User.objects.filter(email__iexact=self.email).exists():
             raise EmailAlreadyKnownException
 
+    def refresh(self):
+        """A simple way to refresh invitation and move expiration date."""
+        self.clean()
+        self.updated_at = timezone.now()
```
qbey: nit: `updated_at` has `auto_now=True`, so a refresh is just a `save`?

src/backend/mailbox_manager/tests/api/invitations/test_api_domain_invitations_refresh.py
```diff
@@ -0,0 +1,112 @@
...
+    with freeze_time("2025-12-16"):
+        invitation = factories.MailDomainInvitationFactory()
+
+    access = factories.MailDomainAccessFactory(
+        domain=invitation.domain, role=enums.MailDomainRoleChoices.OWNER
+    )
+
+    client = APIClient()
+    client.force_login(access.user)
+
```
qbey: I would test the invitation is "expired", just to be sure the time freezing did it's job (yes, just in case)

src/backend/core/models.py
mjeammet: It was my first guess too but upon trying I got `permission denied` for reasons I don't really understand.  Soooo I resorted to update 🤷🏽

PR #1056, opened 2026-02-10
🚸(email) we should ignore case when looking for existing emails

qbey:
## Purpose

Our products mostly rely on email regardless of their case.

Next step would be to normalize email to lower case when storing them in database (or sending them to dimail if not done yet).

## Proposal

- [x] disregard email case when looking for user with specific email address

### DIFF
--- a/src/backend/core/tests/test_models_invitations.py
+++ b/src/backend/core/tests/test_models_invitations.py
@@ -17,6 +17,8 @@ from freezegun import freeze_time
 
 from core import factories, models
 
+from mailbox_manager.exceptions import EmailAlreadyKnownException
+
 pytestmark = pytest.mark.django_db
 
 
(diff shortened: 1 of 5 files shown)


src/backend/core/tests/test_models_invitations.py
```diff
@@ -48,6 +50,20 @@ def test_models_invitations_team_required():
...
+    # Create a user with a lowercase email
+    factories.UserFactory(email="john.doe@example.com")
+
+    # Try to create an invitation with uppercase email - should fail
+    with pytest.raises(EmailAlreadyKnownException):
+        factories.InvitationFactory(email="John.Doe@Example.COM")
+
+    # Try with mixed case - should also fail
+    with pytest.raises(EmailAlreadyKnownException):
+        factories.InvitationFactory(email="JOHN.DOE@EXAMPLE.COM")
```
mjeammet: you swapped the comments / the actual tests

mjeammet (approved):
just a couple lines to unswap and we're good to go !

If you have a bit more time, we could add the dimail part in this PR but i'll do it soon otherwise.

PR #1057, opened 2026-02-11
✨(datagrid) add sort to mailboxes list + mail domain list

elvoisin:
Add sort on all datagrid

### DIFF
--- a/src/backend/pyproject.toml
+++ b/src/backend/pyproject.toml
@@ -26,7 +26,7 @@ readme = "README.md"
 requires-python = "~=3.14.2"
 dependencies = [
     "Brotli==1.2.0",
-    "PyJWT==2.11.0",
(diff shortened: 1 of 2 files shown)


mjeammet (commented):
Not qualified to judge the code but tested locally and it's great.

Did you swap first name and last name in display, tho ? In prod it's "first name last name" and when tested this PR locally, it's the other way around. Was there a reason or it's a simple mistake ?

qbey (approved):

elvoisin:
> around



> I'm not qualified to judge the code but tested locally and it's great.
> 
> Did you swap first name and last name in display, tho ? In prod it's "first name last name" and when tested this PR locally, it's the other way around. Was there a reason ?

It makes more sense to sort by last name, so I think it should be displayed first.

PR #1075, opened 2026-02-25
🐛(i18n) fix missing translations for status tag labels

lslaroche:
(made with ai, review carefully)
Status labels in the Tag component ("enabled", "action required", etc.) were not translated because `t(status)` used a dynamic variable that i18next-parser cannot extract.

Replace with a static `translations` map using literal `t()` calls so the keys are properly detected during extraction and sent to Crowdin for translation.


lslaroche:
à merge

PR #1077, opened 2026-02-26
🚸(frontend) prioritize mail domains as default landing page

lslaroche:
(made with ai, review carefully)
Swap the redirect priority so that users landing on the root page
are directed to /mail-domains/ instead of /teams/ by default. This applies both to the permission-based checks and the fallback using the TEAMS_DISPLAY feature flag.


lslaroche:
> At least 1 approving review is required by reviewers with write access
 maybe @elvoisin @mjeammet ?

PR #1085, opened 2026-03-16
Fix/update logo images

lslaroche:


lslaroche:
@mjeammet to check
i tried to remove claude from contributors, add commit messages, and changelog entries. if i did something wrong tell me so i can improve

PR #1100, opened 2026-03-31
(dimail) replace passwords with connection link

mjeammet:
## Purpose

Stop sending passwords and send login links instead. Both upon mailbox creation and upon requesting login help.

## Proposal

Description...

- [x] Link is sent upon mailbox creation
- [x] Link is sent upon calling endpoint mailboxes/mailbox/send_link/
- [x] Doc updated
- [x] (front) Remove "reset password" button and replace it by "login link"

Code for password-based logic remains, for other instances (hello 👋🏽)

image

## Improvements

It would be great to add some explanation for people expecting a password.

### DIFF
--- a/docs/interoperability/dimail.md
+++ b/docs/interoperability/dimail.md
@@ -2,15 +2,13 @@
 
 ## What is dimail ? 
 
-The mailing solution provided in La Suite is [La Messagerie](https://webmail.numerique.gouv.fr/), using [Open-XChange](https://www.open-xchange.com/) (OX). OX not having a provisioning API, 'dimail-api' or 'dimail' was created to allow mail-provisioning through People.
+The mailing solution provided in LaSuite is [Messagerie](https://webmail.numerique.gouv.fr/), using [Open-XChange](https://www.open-xchange.com/) (OX). OX currently does not provide a provisioning API, hence 'dimail', meant as an interface between Messagerie and outside parties such as People.
 
 Dimail-api and its documentation can be found [here](https://api.osprod.dimail1.numerique.gouv.fr/docs#/).
 
 ## Features
 
-### Domains
-
-#### Domain creation 
+### Domain creation
 
 Upon creating a domain on People, the same domain is created on Messagerie and will undergo a series of checks. The DNS configuration provided by Messagerie can be found in domain administration menu.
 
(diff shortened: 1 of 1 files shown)


src/backend/mailbox_manager/utils/dimail.py
```diff
@@ -351,6 +370,31 @@ def _send_mailbox_related_email(
                 recipient,
             )
 
+    def get_code(self, mailbox):
+        """Get a login link from dimail."""
+        if not mailbox.secondary_email or mailbox.secondary_email == str(mailbox):
+            raise exceptions.ValidationError(
+                "Logging in with a link requires a secondary email address. Please add a valid secondary email before trying again."
+            )
```
qbey: This is not the job of this helper to check that (because we could also want to request a code from the frontend, to send it via Tchap or any other messaging product).

src/backend/mailbox_manager/api/client/serializers.py
```diff
@@ -63,20 +63,23 @@ def create(self, validated_data):
...
+            # send confirmation email
+            client.notify_mailbox_creation(
+                recipient=mailbox.secondary_email,
+                mailbox_data={
+                    "email": mailbox_data["email"],
+                    "password": mailbox_data["password"],
+                },
+                issuer=self.context["request"].user,
+            )
+
```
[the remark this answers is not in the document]
mjeammet: OK, I'd wondered about that before your arrival and had read that here was the right place for business logic but I can't find the source anymore. Where would be the best place, then ?

src/backend/mailbox_manager/api/client/serializers.py
```diff
@@ -63,20 +63,23 @@ def create(self, validated_data):
...
-            else:
+            if not mailbox.secondary_email:
                 logger.warning(
                     "Email notification for %s creation not sent "
                     "because no secondary email found",
                     mailbox,
                 )
 
+            # send confirmation email
+            client.notify_mailbox_creation(
```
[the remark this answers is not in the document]
mjeammet: By the way, a while ago, we added dimail-related notifications methods (such as `notify_mailbox_creation`) to the dimail client/util. I wonder if this is the right place for these methods, belonging to the business logic. I was tempted to move them to the serializers before your above comment.  WDYT ?

src/backend/mailbox_manager/api/client/serializers.py
```diff
@@ -63,19 +64,25 @@ def create(self, validated_data):
...
                     "because no secondary email found",
                     mailbox,
                 )
+            else:
+                mailbox_data["link"] = (
+                    f"{client.API_URL}/code/{client.get_login_code(mailbox)}"
+                )
+
+                if not settings.SEND_MAILBOX_PASSWORD:
+                    mailbox_data.pop("password")
```
qbey, line 79: Useless because you already check `settings.SEND_MAILBOX_PASSWORD` value in `notify_mailbox_creation`

src/backend/mailbox_manager/utils/dimail.py
```diff
@@ -290,26 +290,53 @@ def _raise_exception_for_unexpected_response(self, response):
...
-        and send new mailbox information.
+        Send email to confirm mailbox creation, provide mailbox information
+        and connexion url.
         """
+
         title = _("Your new mailbox information")
         template_name = "new_mailbox"
+
+        if settings.SEND_MAILBOX_PASSWORD and "password" in mailbox_data:
+            template_name = "new_mailbox_password"
```
qbey, line 301: I don't see any test which checks this behavior.

src/backend/mailbox_manager/api/client/serializers.py
mjeammet: I wanted to pop the password here because I thought that, maybe, the least we handle this specific data, the better ? By not re-sending it to the dimail client or to mailing components, we might reduce the "attack surface", WDYT ? @qbey

PR #1124, opened 2026-05-07
⚰️(auth) remove oauth2 and IdP-related features

mjeammet:
## Purpose

remove unused identity provider-related features

## Proposal

Description...

- [x] remove mailbox_oauth2 tables
    this required removing models inheritance from `oauth2_provider` and changing oauth2-related settings to self
- [x] remove every feature related to mailbox-oauth2
- [x] remove tests
- [x] remove doc and add changelog
- [x] update mailbox tests impacted by changes


src/backend/mailbox_oauth2/migrations/0001_initial.py
qbey, line 1: You cannot simply remove the migration: you need to keep the application, add a migration to remove the tables, make a release, then after the release has been deployed in production, remove the application (and migration files of course). Otherwise the tables will stay in the Postgres for ever ^^
Also, you will probably need to do the same for the tables created by oath-toolkit: but in this case you will have to create a manual migration to remove the tables (as you cannot edit the library code ^^)

qbey (commented):
Nice cleaning! 

I left a comment because there is a trick to also clean the database in production.

src/backend/mailbox_oauth2/migrations/0001_initial.py
mjeammet: Ahah yes, removing the migration was a careless mistake on my part but I didn't realize I had to keep the app. Thanks for the tuto, i'll try and do that.

mjeammet: Well, I've been struggling !

`oauth2_provider` wouldn't let me drop the models in `mailbox_oauth2`. I think I've managed to find a way, transiting through normal models, for oauth2 to stop bothering me. 

EDIT : nope, still broken !

mjeammet:
So, the models are badly entangled with `oauth2_provider`'s configuration, maybe because of swappable models (?). In any case, 95% of the dead code removed in this PR has been successfully removed and I'd like to merge it. So, I'm leaving the models for now and will be opening an issue in hope that future me or @qbey or @jmaupetit will find the courage to tackle this issue, someday.

mjeammet:
As discussed IRL, @qbey is OK for me to merge it as is and we'll come back for these damn models soon™️.

PR #1165, opened 2026-09-07
🔒️(security) match organization email domains exactly

hadrienblanc:
## Purpose

Organization auto-assignment at login matched the email domain with `icontains`, a substring match. A user controlling a domain like `ontpellier.fr` was silently attached to an organization registered with `ville-montpellier.fr`. As a member of the victim tenant, the user could read its whole user directory, all member profile contacts and organization details.

## Proposal

- replace the two `icontains` filters in `OrganizationManager.get_or_create_from_user_claims` with exact array containment
- keep the matching case-insensitive, since email domains are: lowercase the incoming domain before matching and on organization creation
- add data migration `0018_organization_domain_list_lowercase` for existing stored domains
- add two regression tests in `core/tests/authentication/test_backends.py`:
  - a substring domain does not join the organization, a separate one is created
  - an uppercase domain still matches its organization

### DIFF
--- a/src/backend/core/migrations/0018_organization_domain_list_lowercase.py
+++ b/src/backend/core/migrations/0018_organization_domain_list_lowercase.py
@@ -1,0 +1,22 @@
+from django.db import migrations
+
+
+def lowercase_domain_list(apps, schema_editor):
+    """Domains are case-insensitive: store them lowercased so exact matching works."""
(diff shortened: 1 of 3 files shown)


mjeammet (approved):
Hi ! Good looking out ! Thanks for your PR.
Looks good to me :)

Could you just add a line on the changelog, please ?

